PR #2, opened 2020-11-13
complete restructuring to snakebids app

akhanf:
- still a work in progress, but should be a "mostly" working version now!
- relies on https://github.com/akhanf/snakebids for bids grabbing and CLI (which is still also under construction)
- includes https://github.com/jordandekraker/Hippocampal_AutoTop as a submodule for the core unfolding code
- pre-processing for T1w, T2w, and dwi (currently dwi processing limited to data that can be run through topup, e.g. multiple phase encode repeats)
- please add more feature requests as github issues!


akhanf:
also forgot to mention, workflow is not yet packaged into a single container yet -- will work on that

PR #9, opened 2020-11-20
Manual segmentation as input

akhanf:
- uses the --segT1w or --segT2w flag to use manual segmentations as inputs
- requires use of --derivatives option to specify the bids derivatives folder where dseg images can be found
- some other file/folder renaming too
@royhaast let me know when you're happy with the last few tweaks and we can merge in

### DIFF
--- a/config/snakebids.yml
+++ b/config/snakebids.yml
@@ -64,6 +64,17 @@ pybids_inputs:
       - acquisition
       - run
 
+  seg:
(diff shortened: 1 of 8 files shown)


akhanf:
forgot to mention, is related to #3

royhaast:
Yep, I'm happy! So feel free to merge

Issue #8, opened 2020-11-16
repackage singularity


jordandekraker:
We're bootstrapping a container we're not using (from khanlab/hippocampa_autotop) which is bootstrapping another container (khanlab/autotop_deps) and still we don't have snakemake & snakebids built in. We should just combine all of these into a single container.

akhanf:
That dockerfile is no longer used, the snakemake workflow is currently dependent on the autotop-deps container, and the prepdwi container, and uses locally-installed python and matlab.

Since we are just running from source, right now it is easier to avoid having a single container so we don't have to rebuild a container code changes. Actually the more I think about it, putting all the dependencies in one container is not only inefficient but it may be very tricky since we're using old (niftynet) and new (snakemake) python code..  I think having container management built-in to snakebids is a better idea - i.e. having the script pull any containers needed when installing, or referencing from a shared container folder, etc..)..  

Here are some instructions to get your python environment set-up on graham to run hippunfold:

1. create a virtualenv and activate it:
```
mkdir $SCRATCH/hippdev
cd $SCRATCH/hippdev
module load python/3
virtualenv venv
source venv/bin/activate
```
2. clone the source repos (so you can make/pull changes easily, or change to branch):
```
git clone http://github.com/akhanf/snakebids
git clone --recursive http://github.com/khanlab/hippunfold
```

3. install snakebids using pip, with the -e option (for development mode):
```
pip install -e ./snakebids
pip install -e ./hippunfold
```

Now hippunfold will be installed for you and can run with:

hippunfold  <args here> 

Any containers used are included in the hippunfold workflow, and if in khanlab group on graham, will already be good to go..  If you log out, you just need to re-activate the virtualenv to start again 

If you ever want the latest code, can just pull it:
```
cd snakebids
git pull
```

or if you need a branch, can: `git checkout <name of branch>`

jordandekraker:
realized the specific error message I was getting was just because I needed to update snakebids in my environment, but cases where no T1w is available still seems like an issue to me.

akhanf:
alright finally got around to putting things together in a single container (all but dwi workflow at least, which is still experimental anyways..), can see details in #16

PR #16, opened 2020-12-14
Packaging for pip and docker container

akhanf:
1. Refactored files/folders and packaging to put everything in a hippunfold module. Now `pip install <repo dir>` will install hippunfold binary, and will copy all the workflow and autotop files to the python site-packages. Note that since the total file size is large (niftynet networks, atlases, etc), we currently won't be able to push to PyPI (max 60mb) this way, so in future should prune these away and download during or before workflow to get simple `pip install hippunfold` to work..

2. Compiled MCR for autotop, ironing out any errors (paths relative to m files), currently pointing to akhanf-dev2 branch on hippocampal_autotop

3. Updated Dockerfile for standalone multi-platform version, addresses #8 

- [x] test T2w single container run with Docker using autotop_deps:v0.2 base on my mac
- [ ] merge akhanf-dev2 branch on hippocampal_autotop once tested


akhanf:
merging this in, will update the autotop submodule later

PR #24, opened 2021-01-18
nnU-Net and some not so minor refactoring

akhanf:
* uses nnUnet for CNN Unet segmentation instead of niftynet
* currently only have T2w model trained, training for new data can be done with https://github.com/akhanf/nnUNet_smk
* models are currently copied to my dropbox for http download, I am open to other suggestions!
* created download_models.py script to download nnunet model tars, can be run as hippunfold_download_models
* nnunet segmentation is passed on to autotop as a "manual" segmentation, in order to override niftynet
* refactoring of autotop.smk into subfields_vol.smk and subfields_surf.smk, and removed redundant/unused files from output

To do:
 - [x] test run from a clean venv
 - [x] train models for T1w, ~~dwi, +~~ @jordandekraker: let me know which models you are planning on training.. for HCP1200 models, we should try to use the exact same set of subjects for each model
 - [x] upload models to web (dropbox currently)


jordandekraker:
Running test from clean environment: seems to work.
NOTES: with `--modality T1w` or other modalities it return an error in rule `nnunet.smk`. This is fine once we generate nnunet models for all supported modalities, but for now perhaps this should just return a warning and skip nnunet rather than an error. Also, the labelmap file in the corobl folder (usually `niftynet_lbl.nii.gz`) is instead named `manual_lbl.nii.gz`, which is a bit confusing but otherwise lookds great!

As for training new models, I want to run:
-HCP1200-T2w
-HCP1200-b500
-HCP1200-T1w
-HCP1200-1mmT1w
Unfortunately the training data I have for these do not have matched participants. I'll go ahead and make a subset matched all participants.
-7T-TSE (roughly 0.4x0.4x1.0mm) from Berron et al and from Roy&Mohamed (currently n=32, but they're all missing dummy label 4)

akhanf:
Yeah, I was planning for this PR to replace all the niftynet, so I wasn't planning on keeping both types of files in the workflow..  I agree the naming inside autotop isn't the most intuitive as manual_lbl, but can perhaps live with it for now, as it may get replaced in the future.. 
 
Some comments on models:
- we should actually call the b500 model HCP1200-b1000 instead, as it is trained on b1000 data  (I only had it called b500 because of it's intended use for a b500 (hipp dwi) dataset)..
- The 1mmT1w dataset shouldn't be needed going forward, as the T1w model should suffice, since blurring augmentation is performed in nnunet..
- I would suggest training the 7T model on just the Berron data for now (leaving out the other dataset with missing label), since that missing label would cause more problems than help I think.. 
- For the 7T model (or any dataset with high-res inplane, thicker slices) we may actually have better performance if we don't resample to isotropic, as I just remembered nnunet does things differently for anisotropic scans (e.g. in setting up with filter sizes) -- maybe that's something we can explore on the side: I am thinking of a different pre-processing pipeline for those kinds of scans, that would skip the corobl registration/resampling, and just trim to fit a standard box instead..

jordandekraker:
Sounds good.

A few more things on the TSE data: both datasets are missing label 4. Also, the segmentations are in 0.3mm iso, and I worry that if we downsample them to 0.4x0.4x1.0mm then the unfolding may fail (eg. holes in the dark band, etc).
0.3mm wasn't necessarily chosen to suit the incoming data, but rather the resolution at which the unfolding coordinates work well

akhanf:
Ah, ok -- 0.4x0.4x1 doesn't seem all that anisotropic actually, so may not
help much anyways to keep it anisotropic.. Fingers crossed that it just
works fine the first time around!

jordandekraker:
So I'm just trying the nnunet model that you trained on the TSE data. There is no T1w to go with it, so as in [this issue](https://github.com/khanlab/hippunfold/issues/12), I applied a temporary workaround by copying the T2w and naming it T1w for registration to CITI. For all images in space-corobl, it got weirdly stretched along the slice dimension, like this:
hippunfold_nnunet
Funny that its more or less correctly cropped, just stretched. Seems like the atlas registration succeeded, but maybe the slice thickness was not read correctly?

Funnily enough, the segmentation worked reasonably well even on this haha

akhanf:
as per slack discussion with @jordandekraker, added rigid reg cmdline option which fixes the stretching issue (was stretching in affine registration to template)

akhanf:
@jordandekraker your T1w model is in, gonna run a final test with `--modality T1w T2w` and if all is good will merge this in!

PR #29, opened 2021-01-29
minor clarifications

jordandekraker:
If you disagree on whether this is clearer, feel free to revert!

### DIFF
--- a/.github/workflows/python-testing.yml
+++ b/.github/workflows/python-testing.yml
@@ -19,6 +19,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+      with:
+        submodules: True
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v2
       with:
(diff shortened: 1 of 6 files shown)


akhanf:
these look good to me, we can keep this branch open for additional corrections or additions

jordandekraker:
Just trying out the new file outputs and thought maybe it would be good to also include `/anat/sub-{subject}_hemi-{LR}_space-cropT1w_desc-preproc_T2w.nii.gz`, so its available as an underlay for corresponding segmentation images in `/seg_T2w/` (eg. in ITK-SNAP where segmentation must have the same size as the underlay)

akhanf:
That image is actually already included but in the seg_T2w folder -- I
wanted to put it in anat folder, but right now the crop ref is based on the
segT2w.. we could move it to the anat folder but would need something else
in the filename, e.g. from-T2w, which is cumbersome..

PR #42, opened 2021-03-22
UnfoldingforDiffusion

Bradley-Karat:
Adding unfolding for diffusion from Uzair Hussain


akhanf:
will need some work, let's use the new branch graddev instead

PR #43, opened 2021-03-22
graddev integration

akhanf:
This PR is for creating the graddev file from the coords to allow us to do dwi fitting in unfolded space.

@Bradley-Karat let's use this branch to implement the new functionality.  We can walk through this together as it is the best way to learn.. 

So far I have put the files you gathered into the `hippunfold/tmp_from_uzair` folder.. 

First step for you would be to create a local virtualenv to develop in, e.g. see: https://hippunfold.readthedocs.io/en/latest/getting_started/installation.html#setting-up-a-dev-environment-on-graham

Then try getting the `example.py` code working. We may need to add some packages to the requirements for that too.. 

Then we can figure out what functionality we want to add into the workflow and how..

### DIFF
--- a/hippunfold/tmp_from_uzair/README.md
+++ b/hippunfold/tmp_from_uzair/README.md
@@ -1,0 +1,8 @@
+# unfoldingForDiffusion
+
+In example.py is a typical way you would use the unfolding class.
+
+You can use loadCoordinates to load the coordinates (here you will likely need to go in and modify what you want your trailing suffix to be for each coordinate file, right now it is set to U.nii.gz, V.nii.gz and W.nii.gz). This is in the coordinates.py file under the function loadCoordinates, you can also supply a prefix here. It is designed to concatenate and load. When coordinates are loaded it will do all neccessary computations including grad_dev which DOES NOT use acquisition graddev at the moment and is set to rotation only.
+
+Then you may load diffustion and then unfold it. This will populate various unfolding volume related attributes in the unfolding class which you can then save in the folder of your choice. From here on you can fit which ever model you want. Make sure that where ever you save your unfolding volumes you also copy over the bvecs and bvals.
+
--- a/hippunfold/tmp_from_uzair/coordinates.py
+++ b/hippunfold/tmp_from_uzair/coordinates.py
@@ -1,0 +1,520 @@
+import ioFunctions
+import numpy as np
+from scipy.interpolate import griddata, LinearNDInterpolator, Rbf
+from scipy.spatial import KDTree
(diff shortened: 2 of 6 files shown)


akhanf:
To recap here, I've added a rule and script for computing Jacobian with the native-to-unfold image warp.

Just realized however, that we may need the warp in the opposite direction to get the graddev, since this warp actually transforms points/vectors from unfold->native (needs the opposite direction to transform images).  

I'll add a rule to create that composed warp too as I don't have it in the workflow yet, but should be straightforward..

akhanf (commented):
Forgot to mention I also made the Jacobian rule generic

akhanf:
OK getting the composite warps from unfolded to native was more painful than expected, as I found an orientation bug in one of the warps (wasn't being used anywhere else, so didn't catch it before) that was causing problems.. Then I realized the hard way that the ANTS ComposeMultiTransform tool didn't work when composing these maps, but antsApplyTransforms can do the same thing without failing..  

In any case, we have rules to generate the transforms in the other direction now too.. Because of interpolation issues at the boundary of the hippocampus (could try to resolve at some point), will be better for now to use the warps from unfold to corobl (instead of unfold to cropT1w):
```
results/sub-{subject}/sub-{subject}_hemi-{L,R}_from-unfold_to-corobl_mode-image_xfm.nii.gz
```

@Bradley-Karat you can start by perhaps updating the jacobian calculation code to make use of a mask (e.g. like the gradient with nans function Uzair has). Once you have that, then can try the polar decomposition, and finally generating the graddev. For those latter steps, the three of us can chat briefly first. 

Note I also made a change to the jacobian rule to make it fully generic, so we don't need to make new rules when we apply it to different xfms..

akhanf:
> Forgot to mention I also made the Jacobian rule generic

and moved everything related to composing warps to compose_warps.smk

hippunfold/workflow/rules/compose_warps.smk
```diff
@@ -93,4 +92,11 @@ rule compute_warp_jacobian:
     group: 'subj'
     script: '../scripts/compute_jacobian.py'
 
+rule compute_graddev:
+    """ Generic rule for computing graddev from jacobian"""
+
+    input: '{prefix}_jacobian.nii.gz'
+    output: '{prefix}_grad_dev.nii.gz'
```
akhanf, line 99: minor point, but should rename suffix to `graddev` instead of `grad_dev` to be bids compliant (suffix should only have letters/numbers)

akhanf:
warp composition is incorporated in #143

PR #61, opened 2021-05-13
fix in create_spec_file

akhanf:
was setting all to INVALID since it wasn't checking for hemi in wildcards properly..


akhanf:
note: bug was introduced in #41

PR #64, opened 2021-05-19
No matlab

akhanf:
@jordandekraker making this draft PR so I can bring it up to date with master, and translate the warps/gifti code too. I've integrated everything else so far and tested only with dry run, will try running for real soon..


akhanf:
This should be ready to merge now:
- replaces all matlab/mcr with python
- shape injection uses greedy (much faster)
- laplace coord
- default behaviour will tar up each subject's work folder after completion (--keep_work to prevent)

Some remaining things to fix in a later PR:
- midsurf visualization disabled right now, failure could be due to fov of surfaces (?) need to investigate
- documentation / install instructions could be updated

PR #74, opened 2021-06-24
Custom density unfold templates

akhanf:
This PR adds a workflow to build unfold templates with a range of total vertex count, and variable spacing to ensure vertices are evenly sampled (on average) in the native space.

Tasks:
- [x] create workflow to generate new templates
- [x] add new templates to resources
- [x] update workflow to make use of these templates


akhanf:
If you want to see an example of these warped to a subject, they are here:
https://github.com/khanlab/hippunfold/tree/custom-density/gen_unfold_template/example_transformed

and the new unfold templates are here:
https://github.com/khanlab/hippunfold/tree/custom-density/hippunfold/resources/unfold_template_avgarea

(note the naming of files in the example is by target vertex area, whereas it is density in the unfolded template, but they have 1-1 correspondence (e.g. 0.1mm^2 is 11k vertices)

jordandekraker:
Beautiful implementation, Ali! Seems to be working well on my end, just a couple of minor issues

bug: when I load one of these resampled surfaces in Freeview, it crashes when i pan to the edge of the surface (but shows up correctly in the 3D display). I thought this might have to do with the surface exceeding the bounding box maybe? Might just be a Freeview thing though.

suggestion: maybe default number of vertices should be 2k. I think many neocortical atlases have on the order of 32k vertices (eg. conte69 as Boris is using), so 2k would probably be roughly proportional.

suggestion: specifying --output_density 32k could use the original 126x254 mesh

akhanf:
> Beautiful implementation, Ali! Seems to be working well on my end, just a couple of minor issues
> 
> bug: when I load one of these resampled surfaces in Freeview, it crashes when i pan to the edge of the surface (but shows up correctly in the 3D display). I thought this might have to do with the surface exceeding the bounding box maybe? Might just be a Freeview thing though.
> 

Ah not sure what might be the problem there -- do you see the problem if you load one of the new flat unfolded meshes too (ie from here: https://github.com/khanlab/hippunfold/tree/custom-density/hippunfold/resources/unfold_template_avgarea)? Wonder if it is the constrain to bounding box step that moves the vertices into the bounding box, though that is the same as it was before.. I haven't seen any issues when viewing in wb_view.

> suggestion: maybe default number of vertices should be 2k. I think many neocortical atlases have on the order of 32k vertices (eg. conte69 as Boris is using), so 2k would probably be roughly proportional.
> 

2k was from a target surface area of 0.5mm^2, and has a mean vertex spacing of ~0.9mm.  I wasn't considering mean vertex spacing when I generated these, but that might be a better objective than surface area, since can more easily relate it to image resolution. At 0.9mm I think this would be fine for 1mm MRI data, but it is a bit undersampled for higher-res data..  I actually think it would be nice to have the following three levels by vertex spacing:

~0.5mm
~1mm 
~2mm

Right now we have approx (vertex spacing ~ num vertices)
0.4mm ~ 11k
0.9mm ~ 2k
1.2mm ~ 1k
1.8mm ~ 500
2.3mm ~ 300

So perhaps we can instead have three: 10k, 2k, 400

> suggestion: specifying --output_density 32k could use the original 126x254 mesh

I like that idea, and will be easy to achieve by just copying the old mesh over and calling it 32k.

royhaast:
This is awesome. I just checked the template surfaces and they look good. I think this will definitely increase possibilities for joint hippocampal-cortical analyses. 

I agree that vertex spacing is probably the most intuitive for users to decide which resolution would be most appropriate for them based on their data. For example, I can imagine that studies interested in high-res imaging want to use the smaller vertex spacing (i.e., 0.5 mm) while in case of fMRI data 1 or even 2 mm might be preferred.

jordandekraker (approved):

PR #76, opened 2021-06-29
Dice qc

jordandekraker:
This branch provides an additional QC metric by comparing hippocampal segmentations obtained from nnunet to those obtained by a much simpler method: f3d registration to the CITI168 template. This considers only the entire hippocampal volume. A dice score <0.7 is likely indicative of an issue with the nnunet segmentation.


jordandekraker:
I believe as of 7a568f15da79e739b58b6d4450ee682bab374e29 all the code is in place, but something still seems to be broken. 

It seems snakemake cannot find the input for rule qc_nnunet_dice, which should be generated by rule qc_nnunet_f3d. When you get the chance, do you think you could have a look, @akhanf ?

akhanf:
@jordandekraker looks like this also includes commits from the other custom_density branch -- that might cause issues when merging things — to keep things clean do you mind creating a fresh branch from master that just has your commits? (ie 1441e61 and 7a568f1)?  Also, I noticed #75 has the same change as is here, so if that's the case can either remove that PR to include the change only here, or keep them separate?

jordandekraker:
Ahh sorry when I created this branch I had commits from various other branches in my path still. I'll try to clean it up.

jordandekraker:
I just reverted the changes from other branches, so the commit history is a bit messier here but it should be easy to see the `Files changed`. Let me know if this is still to hard to work with and I can delete the branch and try again

akhanf:
> I just reverted the changes from other branches, so the commit history is a bit messier here but it should be easy to see the `Files changed`. Let me know if this is still to hard to work with and I can delete the branch and try again

Thanks! This works fine now, and we can squash the commits when the time comes to keep the master history clean.. Will take a look now!

akhanf:
@jordandekraker just made some commits to fix the workflow, should dry-run properly now (haven't tested it further myself)..

The edits that were causing it to not find the files were:
- replaced input functions with string directly (could have also used
  input functions, but would have needed to use the wildcards object for
  string formatting, so this is more compact..) 
- removed wildcard constraints (was Lflip|R) for one of the rules, this was limiting the rule to only be run on Lflip or R, so wasn't able to produce the L output..

akhanf:
made some other cosmetic changes too, details in the commit message. Once you've tested it let me know, and I'll squash and merge it in.

jordandekraker:
Thanks for the help, Ali! Got things working on my end now.

Not sure if it belongs in the PR, but we could also add a group-wide QC report which would include a histogram of dice scores and a list of those with <0.7.

akhanf (approved):
changes look good to me, I will squash and merge in

akhanf:
can add the group-level workflow in another PR

akhanf (approved):

PR #84, opened 2021-07-08
added --hemi option, and updated workflow

akhanf:
- updates to workflow to remove any dual hemi dependencies from single
  hemi outputs
- dlabel and metric ciftis now will just be single hemi if only one hemi
  is specified
- added test cases for this option too

### DIFF
--- a/hippunfold/config/snakebids.yml
+++ b/hippunfold/config/snakebids.yml
@@ -210,6 +210,18 @@ parse_args:
     nargs: '+'
     help: 'Sets the output vertex density for results (default: %(default)s)'
 
+
+  --hemi:
+    choices:
+      - 'L'
+      - 'R'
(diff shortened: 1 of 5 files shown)


jordandekraker:
I can't seem to get it to run on only hemi-R, perhaps I'm missing something. I tried a few combinations of wildcards in the `--path_cropseg` with and without the `--hemi R` flag but it still complains unless an input hemi-L is at least present

akhanf:
Ah, I introduced a bug in one of my later changes, but should be fixed now..

It was using left and right inputs for the ciftis, now it uses a single input function (and unpack()) to get all the inputs for the rule as a dict.. (note: this is the approach you can use for the other PR I commented on)

jordandekraker (approved):

Issue #85, opened 2021-07-09
faster laplace solver through image pyramids


jordandekraker:
@akhanf has some code to do this already, check out his old nomatlab (I think?) branch

This is really only an issue when running with no initialization (ie. skip_template_shape_injection)

akhanf:
I located where the snakefile with multi-res laplace is, it is actually in master already here:

https://github.com/khanlab/hippunfold/blob/master/hippunfold/resources/tpl-upenn/Snakefile

There is a rule to downsample coords:
 https://github.com/khanlab/hippunfold/blob/49a2245d3fbf5b52c3955a7909a3af38d72041f0/hippunfold/resources/tpl-upenn/Snakefile#L26

Run laplace without initialization at the lowest resolution:
https://github.com/khanlab/hippunfold/blob/49a2245d3fbf5b52c3955a7909a3af38d72041f0/hippunfold/resources/tpl-upenn/Snakefile#L37

Then upsample coords:
https://github.com/khanlab/hippunfold/blob/49a2245d3fbf5b52c3955a7909a3af38d72041f0/hippunfold/resources/tpl-upenn/Snakefile#L72

and run Laplace with init coords:
https://github.com/khanlab/hippunfold/blob/49a2245d3fbf5b52c3955a7909a3af38d72041f0/hippunfold/resources/tpl-upenn/Snakefile#L54

The number of levels is defined by setting the wildcard constraints in the level wildcard, e.g. right now lowest res is level 2 (downsampled by 50% twice; level 1 is downsampled once, level 0 is original resolution)..

akhanf:
speed-up by fast marching initialization #88 obviates the need for this right now, so closing this issue

PR #88, opened 2021-07-10
fast laplace via initialization

jordandekraker:
I recycled some old matlab code to get a fast (and good quality) initialization for the lapalce solver. Only dependency is scikit-fmm which is pretty light.
This could be used in conjunction with image pyramids as in https://github.com/khanlab/hippunfold/issues/85, but honestly I think this is fast enough now. 

Tested and working on my end!

### DIFF
--- a/hippunfold/workflow/scripts/laplace_coords.py
+++ b/hippunfold/workflow/scripts/laplace_coords.py
@@ -51,8 +51,8 @@ backward = skfmm.travel_time(phi,np.ones_like(lbl))
 backward = backward.data
 # combine
 forward = forward/np.max(forward)
-backward = -backward/np.max(backward) +1
-init_coords = np.sqrt(np.square(forward) + np.square(backward))
+backward = backward/np.max(backward)
+init_coords = np.sqrt(np.square(forward) - np.square(backward) +1)
 init_coords = init_coords/np.max(init_coords)
 init_coords[idxgm==0] = 0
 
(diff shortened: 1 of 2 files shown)


akhanf (commented):
I don't know if its a great idea to make this combined initialization the new default.. 
The combination is not really principled by just averaging them together.. I think it would be more sensible to have an option to select the initialization type.

jordandekraker:
Could also just keep the forward or the backward. The only problem is the end point from one direction doesn't match the start point from the other, so I wanted the combination to be weighted more heavily towards the start of each march. If that's too complicated then just one of the two marches would still be a good initialization

jordandekraker:
Now that I think about it, if i want to weight their combination, then maybe I should flip them both, square them, flip one back, average them, and then square root the whole thing.

Whatever the combination method is (just using one direction, averaging them, or doing a weighted average) is still a huge improvement over just initializing with a value of 0.5, and so choosing a combination method is a pretty minor change in comparison.

akhanf:
Sorry I may have misunderstood the change when I glanced over it, I thought
you were averaging the injected coords with fast marching coords?

jordandekraker:
Nope, this is only for when there is not shape injection (and therefore no initialization injection).

jordandekraker:
Did some quick testing and just averaging the forward and backward march seems to consistently do best (weighting averaging was getting very complicated). Can probably merge anytime

akhanf (approved):

Issue #87, opened 2021-07-09
gyrification measure depends on output_density


jordandekraker:
This is a bit annoying, we should maybe look more into how wb_command calculates gyrification. The evenly tessellated outputs don't seem to contain anything too interesting here (or maybe its just noisier due to sharper edges?). Perhaps we should calculate on the 32k surface and then downsample to 2k and 7k?

akhanf:
What wb_command computes is a surface area per vertex (mean of neighboring
3 triangles area), which is on average (across the surface) designed to be
constant.

The fact is we are not computing gyrification, we are computing surface
area, which is not really the same thing..

The gyrification index is typically some sort of a ratio between surface
area or length of the curved cortex and surface area of length of a
comparably smooth cortex.

E.g. this is one method: https://surfer.nmr.mgh.harvard.edu/fswiki/LGI

Or another way is to just calculate the local surface area enclosed in a
fixed radius sphere at the vertex; folded areas would be greater than
smooth ones. Note this is different from just performing surface based
smoothing of the SA map (since that would just end up capturing the same
area whether folded or not).

jordandekraker:
Yeah, that's what I figured. I like freesurfer's localGI method, and it would probably produce something similar but a bit different to what we have been doing previously.
Just to flesh out what we had been doing, instead of comparing surface area to a native vs smoothed hippocampus we are instead comparing it to a fully unfolded (or rectangular grid) hippocampus. I still think this is a useful measure, but to call it a gyrification index is a bit different from how other are using the term (maybe we could consider calling it surface distortion or something like that?). I also wouldn't mind just keeping this method, calling it gyrification, and clarifying the calculation in the methods section.
If we did freesurfer's localGI then it might be easier to write our own python function rather than add dependencies.

akhanf:
Yes I agree it is something we can easily just implement rather than depend on something else.. 

The existing measurement (ie surface area on the 32k surface) could perhaps just be called surface distortion as you suggest..

PR #93, opened 2021-07-31
new gyrification as surface area ratio

akhanf:
native over unfolded surface area

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,7 +1,7 @@
 [bumpversion]
 commit = True
 tag = False
-current_version = 0.5.10
+current_version = 0.5.11
 
 [bumpversion:file:hippunfold/pipeline_description.json]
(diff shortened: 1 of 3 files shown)


akhanf:
The gyrification metric is now the native surfarea divided by unfold surfarea, and the surfarea metric is also produced now.. 

Ran on one test subject and seems to work well -- the gyrification maps for the new variable density surfaces now have approx same values/range, so the value now does not depend on the density. 

The 32k gyrification looks a little bit different from the others (maybe an artifact of the vertex density), but is proportional to the old gyrification maps (by a multiplicative constant), and the old maps are still available  if needed for backwards compatibility (they are just called 32k surfarea.shape.gii).. But I think we should just avoid the 32k density maps for surface-based features from now on..

akhanf:
closes #87

Issue #92, opened 2021-07-26
Hanging on rule create_warps


jordandekraker:
This is happening only with very high resolution data (46x46x60um). I tried increasing the resources from 16gb to 96gb memory but still same problem (and in the system monitor it doesn't appear to use this all anyway). I can also see this is using a single core, but I've left it running for a whole week and never got an output. 

I know the interpolation step can be somewhat memory-intensive, so perhaps this is the issue?

jordandekraker:
Looking more into this, it seems each call to scipy's `interpolation.griddata` generates a subprocess that doesn't clear after we evaluate it. In this case each takes ~40gb and a very long time to complete, likely exceeding the resources allocated by snakebids. I opened a branch to try and run and then clear these serially to save memory https://github.com/khanlab/hippunfold/tree/create_warps_resource_manage.

If this makes sense and work then I'll make a PR and we can do some further refining

akhanf:
Yes sounds like it needs more memory than available.. 

But is it getting past the first griddata call? If so, then maybe clearing things up might help, but I suspect it's the first griddata call that is causing it to hang.. 

I doubt there would be much you can do except use a large memory system (can check the griddata code to see what it needs to allocate), try to work with downsampled data for this step, or perform this interpolation on chunks of the image at a time..

jordandekraker:
Now that we identified the real problem (see https://github.com/khanlab/hippunfold/pull/94), closing this issue

PR #94, opened 2021-08-09
script create_warps.py resource management

jordandekraker:


jordandekraker:
Proceeding as discussed in https://github.com/khanlab/hippunfold/issues/92

jordandekraker:
Currently, the max matrix size is set to 200*200*200 = 8M. In space-corobl, matrix size is 128*256*128 = 4.2M. 

Could perhaps set the max matrix size even larger, though in my other test cases this hangs (despite never reaching even close to memory limit). 

Another quick note: its not actually the matrix size that matters but the number of GM voxels (which in my ex-vivo case was 456K after downsampling. In space-corobl this is typically ~80K). This could vary quite a bit depending on the sample, so matrix size is probably a better basis for downsampling.

jordandekraker:
After spending a long time trying to get this step (and the next) to work on 3D PLI and BigBrain data at super high resolution, I realized the issue wasn't memory. I'll leave https://github.com/khanlab/hippunfold/tree/create_warps_resource_manage as is in case we need further memory management (in this case by downsampling) in the future.

The actual issue is outlined here:
https://stackoverflow.com/questions/10886971/alternatives-to-scipy-interpolate-griddata-that-dont-hang-on-aligned-points

I created a new branch with this fix: https://github.com/khanlab/hippunfold/pull/95

jordandekraker:
Closing this PR as it is no longer needed, but may be useful if we need downsampling to manage memory in the future!

PR #95, opened 2021-08-17
No more overlapping points in griddata interpolation in create_warps.py

jordandekraker:
After spending a long time trying to get this step (and the next) to work on 3D PLI and BigBrain data at super high resolution, I realized the issue wasn't memory. I'll leave https://github.com/khanlab/hippunfold/tree/create_warps_resource_manage as is in case we need further memory management (in this case by downsampling) in the future.

The actual issue is outlined here:
https://stackoverflow.com/questions/10886971/alternatives-to-scipy-interpolate-griddata-that-dont-hang-on-aligned-points

The solution is to simply add a tiny bit of noise to the Laplace coords to ensure none of them are perfectly overlapping or perfectly aligned along any one dimension. I'm keeping some of the improvements and debugging from the downsampling code since this step still takes ~1h and its nice to check that it is indeed running.

### DIFF
--- a/hippunfold/workflow/scripts/create_warps.py
+++ b/hippunfold/workflow/scripts/create_warps.py
@@ -34,7 +34,7 @@ coord_pd = coord_pd_nib.get_fdata()
 coord_io = coord_io_nib.get_fdata()
 
 #get mask of coords  (note: this leaves out coord=0)
-mask = (coord_ap > 0) | (coord_pd > 0) | (coord_io > 0) # some points were lost, especially IO
+mask = (coord_ap > 0) | (coord_pd > 0) # some points were lost, especially IO
 num_mask_voxels = np.sum(mask>0)
 print(f'num_mask_voxels {num_mask_voxels}', file=logfile, flush=True)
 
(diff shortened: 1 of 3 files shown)


hippunfold/workflow/scripts/create_warps.py
```diff
@@ -34,7 +34,7 @@ def summary(name, array):
 coord_io = coord_io_nib.get_fdata()
 
 #get mask of coords  (note: this leaves out coord=0)
-mask = (coord_ap > 0) | (coord_pd > 0) | (coord_io > 0) # some points were lost, especially IO
+mask = (coord_ap > 0) | (coord_pd > 0) # some points were lost, especially IO
```
akhanf, line 37: does this end up reducing the number of voxels in the mask? (since it is a logical OR)?

akhanf (commented):
just had one question on defining the mask -- see line comment

hippunfold/workflow/scripts/create_warps.py
jordandekraker: it actually increases the number of voxels in the mask - but only by a handful.

akhanf: ah yea, of course! -- I guess my fault for multi-tasking when I read this!

akhanf (approved):
Looks good to me, feel free to merge this in

jordandekraker:
Closing with one comment that I'd like to address in https://github.com/khanlab/hippunfold/pull/96

There are some NaNs in the create_warps.py resulting from the inability of griddata to extrapolate (see https://github.com/khanlab/hippunfold/pull/95/commits/fc1d2522623afad5f445c77f87c4c05cac001940). Currently the shortcut is to set them ==0, but this can create misplaced vertices in the final surfaces. I aim to fill those sensibly

PR #96, opened 2021-08-20
Switch IO laplace (isopotential) coordinates to isovolume

jordandekraker:
Working with ex-vivo data, I was getting annoyed that the inner and outer surfaces don't extend quite to the edges of grey matter in highly gyrified areas. This is because in the depth of a sulcus, the IO laplace coordinates are all nearly 0. Additionally, even the midthickness surface was quite smoothed.
I experimented with a (nearly) equidistant model and found that it ameliorated the issue (but was somewhat bumpy). 
The isovolume model (https://github.com/nighres/nighres) is actually most valid. We used this for 3D BigBrain and always meant to use it more in MRI but never got to it. I'm hoping now that this will solve my issue with the inner/outer surfaces and improve the validity of laminar analysis for ex-vivo work. 

Will need some testing with both in-vivo and ex-vivo data


jordandekraker:
Can't figure out how to execute my script in python3 using snakemake's `script:` . Maybe there is a way to do it with `shell:` but I can't find it. Worst case scenario we can add nighres to the autotop image (but it needs to be built not jsut pip installed)

jordandekraker:
@akhanf no rush, but any chance you'd be willing to rebuild the autotops container with nighres included? its can't be pip installed, has to be built from source code so it may be a bit of a pain...

akhanf:
great -- working on getting nighres in autotop_deps (some things to workaround since the container is a different OS than the nighres Dockerfile), but should be resolvable.. 

I'll also merge the other documentation changes (#97) into this branch and review all the changes so far.

hippunfold/config/snakebids.yml
```diff
@@ -221,6 +221,10 @@ parse_args:
     nargs: '+'
     help: 'Hemisphere(s) to process (default: %(default)s)'
 
+  --NoIsovolume:
```
akhanf, line 224: couple comments:
 - we should use lowercase for argument names
 - but I think might be better to instead have an option called --laminar_coords_method, that can be either "laplace" or "isovolume" (can add others if implemented too), and can default to "isovolume"

hippunfold/workflow/rules/autotop.smk
```diff
@@ -33,6 +33,31 @@ rule laplace_coords:
...
+    script: '../scripts/prep_isovolume_coords.py'
+
+rule isovolume_coords:
+    input: 
+        outerbin = bids(root='work',datatype='seg_{modality}',dir='{dir}',suffix='outerbin.nii.gz',space='corobl',hemi='{hemi,Lflip|R}', **config['subj_wildcards']),
+        innerbin = bids(root='work',datatype='seg_{modality}',dir='{dir}',suffix='innerbin.nii.gz',space='corobl',hemi='{hemi,Lflip|R}', **config['subj_wildcards']),
+    params:
+        src_labels = lambda wildcards: config['laplace_labels'][wildcards.dir]['src'],
+    output:
+        coords = bids(root='work',datatype='seg_{modality}',dir='{dir}',suffix='isovol.nii.gz',space='corobl',hemi='{hemi,Lflip|R}', **config['subj_wildcards']),
```
akhanf, line 53: since isovol coords are still coords, I think we should rename to be: desc='isovol', suffix='coords.nii.gz'

then the laplace coords can be renameds to: desc='laplace', suffix='coords.nii.gz'

hippunfold/workflow/rules/autotop.smk
```diff
@@ -69,14 +101,20 @@ rule create_unfold_coord_map:
     shell:
         'c3d {input.nii} -cmp -omc {output.nii}'
 
-        
+def get_io(wildcards):
```
akhanf, line 104: is this used anywhere?

akhanf: in any case this would be the place to make use of the config parameter and pick the laminar_coords_method, and would presumably then be called in the create_warps input

hippunfold/workflow/rules/common.smk
```diff
@@ -67,6 +67,15 @@ def get_final_coords():
                 space=config['output_spaces'],
                 allow_missing=True)
 
+def get_final_isovol():
```
akhanf, line 70: shouldn't need this if it is already an input to create_warps..

akhanf: with the name change to use the `desc` tag, we could perhaps just use get_final_coords() to include the various methods

hippunfold/workflow/rules/resample_final_to_crop_t1.smk
```diff
@@ -70,6 +70,19 @@ rule resample_coords_native_crop:
         'ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS={threads} '
         'antsApplyTransforms -d 3 --interpolation NearestNeighbor -i {input.nii} -o {output.nii} -r {input.ref}  -t [{input.xfm},1]' 
 
+rule resample_isovol_native_crop:
```
akhanf, line 73: this rule could be left out if we instead update `resample_coords_native_crop` to add `desc` wildcard

akhanf (changes requested):
Haven't tried running it yet, but just left some comments on the code & file naming. 

Will try to give it a go on some in vivo data and also make sure the new combined hippunfold container build works fine.

hippunfold/workflow/rules/resample_final_to_crop_t1.smk
akhanf: fixed

hippunfold/workflow/rules/common.smk
akhanf: fixed

hippunfold/workflow/rules/autotop.smk
akhanf: fixed

hippunfold/workflow/rules/autotop.smk
akhanf: fixed

hippunfold/config/snakebids.yml
akhanf: fixed

akhanf (approved):

akhanf:
@jordandekraker I've got things working on my end now, made some changes to the file naming (using `coords` suffix instead of `isovol`) and the optional arg for specifying the laminar coords method. It will generate the `dir-IO` laplace coords file regardless now, but will use the one specified in the argument (default: isovolume) for the actual downstream processing.

One thing I couldn't get to work right however is the logging to file for nighres functions -- this might be a tough nut to crack since I think it is related to the java wrapper, so I'm ok just leaving it as is.. but let me know if you have any suggestions.

The other minor change I was thinking of (for bids consistency) is to rename the `*_innerbin.nii.gz` and `*_outerbin.nii.gz` images -- if they are binary masks, then the proper suffix should be `mask.nii.gz` [see bids spec](https://bids-specification.readthedocs.io/en/stable/05-derivatives/03-imaging.html#masks)

And finally, I was wondering if we should re-consider our naming of the method -- I don't really see any references that use the "isovolume" nomenclature, as they use "equivolume" or "equivolumetric" instead.. Any objections to replacing isovol and isovolume with `equivolume`?

jordandekraker:
Awesome! I'll give it a few test runs right now. 
I thought the nighres functions were printing to log files when I last ranthis. I'll check again.
Changing the filenames is perfectly fine with me. `_innerbin.nii.gz` could perhaps be `_desc-SRLM_mask.nii.gz` while `_outerbin.nii.gz` could be `_desc-wholeHipp_mask.nii.gz`? These are just temporary files so I didn't think too much about them
Happy to switch isovolume to equivolume too.

jordandekraker:
A few quick updates: I like the changes you made to keep the code simpler & minimize the number of new rules. The nighres code is still not saving into a log file, but it does actually print to terminal like the nnunet or ants outputs do, so I think I'm fine with that. I'll commit to switching isovolume to equivolume and other file names, and after that I expect we'll be good to merge!

akhanf:
sounds good!

jordandekraker:
Tested and working well on my end. 

Before we merge, are there any additional things we want to add before v0.6?
I plan to add more to the tutorial and include a pretty extensive set of example images for all readthedocs sections. I think this will be a basis for a demo at the upcoming BigBrain workshop. I'll probably do this gradually though, and it won't affect any processing.
The `work/` directories are pretty large, so I'm glad we're zipping them by default. I don't know if there's much else I would change about that. 
Anything else we should consider?

akhanf:
I think we can just merge this and increment to 0.6.0 at this time (btw, I should be able to do that by including `#minor` in the commit message with the bumpversion](https://github.com/marketplace/actions/github-bumpversion-action) github action we have set-up, then we can add in additional documentation etc as later patches, i.e. 0.6.x 

Creating a github release is what triggers the tagged PyPI and docker hub releases, so we can always just hold off on doing that until we are ready. That version number would be 0.6.x instead of 0.6.0 but that's fine.

PR #97, opened 2021-08-24
Tutorials & examples

jordandekraker:
Will add a page for common use-cases, breakdown of flags, and will also add images to the outputs page


akhanf:
looks like you also have some isovolume commits here -- was that intentional?

akhanf:
Also, I just enabled the readthedocs on this branch too, so you can have a look at the rendered html after any commit: 
https://hippunfold.readthedocs.io/en/tutorials/

PR #105, opened 2021-09-10
Cropseg patch

### LINKED ISSUE #104: cropseg with only one hemi
If only one hemisphere `_dseg` file is present, then it should run only that without having to specify the `--hemi` flag. In this case, even when I specify the `--hemi` flag it still returns a weird error where it can't find the missing hemisphere in `rule copy_to_results`. It only works if I use `--path_cropseg` and include the {hemi} wildcard there. 
I think this issue depends on whether `hemi` is specified in the config or as a {wildcard}, but I'm not sure how to solve it.

akhanf:
Closes #104.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,7 +1,7 @@
 [bumpversion]
 commit = True
 tag = False
-current_version = 0.6.0
+current_version = 0.6.1
 
 [bumpversion:file:hippunfold/pipeline_description.json]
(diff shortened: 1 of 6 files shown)


akhanf:
@jordandekraker made some corrections that fix things, along with a minor update to docs. Can merge this once you verify this works with your own data too.

jordandekraker:
Yup, working on my end!

PR #110, opened 2021-09-23
adds coords on surfaces, & pre-smoothing for curv

akhanf:
Unfolded coords now generated as a surface metric too, is especially needed if using surfaces other than 32k.

Also added smoothing to mid-surface prior to curvature calculation (thanks @Bradley-Karat for testing out parameters for this step)

Unrelated: Fixed config yml issue that always added rule 'all', even if target rule specified from command-line

- [x] initial testing of the rules
- [x] run entire workflow on a subject and examine results


akhanf:
Seems to be working fine, so merging this in

Issue #113, opened 2021-11-03
qc_reg_to_template failing 


akhanf:
The nilearn visualization in the rule qc_reg_to_template:
https://github.com/khanlab/hippunfold/blob/5c4c4be5689899c79151462ebabaa84e036ced4a/hippunfold/workflow/rules/preproc_t1.smk#L56

seems to be failing now -- my guess is that something changed in nilearn, since we aren't pinning to a specific version of it, but I haven't investigated this in any detail..

Will try to do this some time soon.. Also is a good reminder to probably deal with dependencies in a better way, e.g. we have moved some other projects to use poetry for this, can do the same here too.

akhanf:
I've disabled the outputs of this rule for now to un-break things until we have a fix..

PR #114, opened 2021-11-05
temporarily disable the vis regqc rule

akhanf:
seems to be failing, disabling it until we can fix it..


akhanf:
related to #113

PR #115, opened 2021-11-10
pin docutils<0.18

akhanf:
should fix readthedocs failed build


akhanf:
https://github.com/readthedocs/readthedocs.org/issues/8616

PR #116, opened 2021-11-12
began new unfolded ref space for DG

jordandekraker:
This still needs to regenerate the files in resources/unfold_template_autotopDG

### DIFF
--- a/hippunfold/config/snakebids.yml
+++ b/hippunfold/config/snakebids.yml
@@ -293,23 +293,42 @@ hippdwi_opts:
   bbox_y: '198 453'
 
 unfold_vol_ref:
-  dims:
-    - '256'
-    - '128'
-    - '16'
-  voxdims:
-    - '0.15625'
-    - '0.15625'
-    - '0.15625'
-  origin:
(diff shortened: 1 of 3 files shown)


jordandekraker:
@akhanf would you be able to rerun https://github.com/khanlab/hippunfold/blob/master/gen_unfold_template/Snakefile to generate the files in https://github.com/khanlab/hippunfold/tree/dentate/hippunfold/resources/unfold_template_autotopDG using the reference space generated in this branch? You have the reference surfaces (UR100 from hippunfold v0.5.1). Should also make sure this branch is up-to-date with `dentate`

akhanf:
yep should be able to do that some point this week

akhanf:
I'm a bit confused with the branching -- do you want to make use of the latest commits in the `dentate` branch? If so, we should just merge this branch into that..

jordandekraker:
Yeah i just made this branch so i could test stuff in the other branch without these changes stopping it. I'm pretty much satisfied with things in the other branch so we could merge them now

PR #117, opened 2021-11-25
Cosine_rep to generate smooth, simple nativespace surfaces

jordandekraker:
This is for general visualization purposes (ie. visualization in native space when neither subject-specific nor a group average is appropriate)

TODO: consider transforming to space-MNI152


jordandekraker:
These canonical surfaces might be confused with real subject surfaces if they are in space-MNI152, so instead I manually rotated to corobl and translated to the origin

PR #120, opened 2021-12-01
qc fixes

### LINKED ISSUE #113: qc_reg_to_template failing

akhanf:
Some updates for qc rules:
- closes #113  (at least I can't reproduce that error anymore..)
- adds back the surf qc pngs (needed density wildcards)
- puts dseg qc into a single subplotted png, instead of 3 separate ones
- uses dim=-0.5 for plotting instead of -1 (I was getting very dark images occasionally, this makes it more stable)


jordandekraker (approved):
Looks great, and it worked in my test case!

We may want to add a .html file combining all of these at some point, but that's maybe a future addition

PR #125, opened 2021-12-06
Use x.x.x-pre.x versioning

kaitj:
This patch fixes an issue with `poetry` conflicting with `x.x.x-prerelease.x` versioning (see [here](https://github.com/akhanf/snakebids/issues/86)) by using `x.x.x-pre.x` versioning instead.


akhanf:
closing without merging this since is now merged in with #132

PR #130, opened 2021-12-18
adds rules for composite warps from unfold

akhanf:
- moves all compose warps rules to separate rules file
- adds compose warp rules, and update to target rules

TODO:
- [ ] replace hacky orientation fix
- [ ] testing on real data


akhanf:
oops rebasing this caused some issues, going to close this and start a new PR

Issue #122, opened 2021-12-03
installation dependencies


sheyma:
```
$ cd hippunfold
$ python3 setup.py build
$ python3 setup.py install --user

```

installation is uncessfull, unless these dependencies are installed:

pygraphviz
extension-helpers
astropy
pillow>=8.3.2

akhanf:
Thanks for picking that up!

PR #131, opened 2021-12-18
replaces setup.py with pyproject.toml and poetry

akhanf:
- should fix dependency issues
- still works with pip install
- however for development-mode (pip install -e ) you will need to use:
poetry install (and install poetry beforehand)
- updates documentation to replace pip install -e with pip install from
pypi


akhanf:
closes #122

PR #137, opened 2021-12-18
single modality at a time

akhanf:
- change modality from a list to a string
- removes modality wildcard from all (except work tar file)
- changes seg_{modality} and surf_{modality} to seg_ and surf_ (allowing
for further refactoring)
- T2w workflow updated and dry-run tested, others are WIP


akhanf:
ok this looks good to review and merge in -- I didn't do any other refactoring here (except for renaming `seg_{modality}` and `surf_{modality}` to `seg` and `surf`.

jordandekraker:
This might be a good place for one more thing - perhaps we should change the default modality to `T1w`? This seems to be most common plus in many cases it seems to work better

akhanf:
Yeah I agree, I think defaulting to T2w doesn't make sense anymore.. 

I am thinking we should just make `--modality` a required argument and not have any default at all. That avoids any surprises in changing the default, and is probably safer since we are no longer allowing multiple different modalities in the same folder..

jordandekraker:
Ah didn't realize you updates the name. `warps.smk` will need `'seg_{modality}'` replaces with just `'seg'`

jordandekraker (approved):

akhanf:
ya I updated those already with a recent merge of master.. I'll fix the remaining conflicts now (after merging in neonatal) then merge this in

Issue #128, opened 2021-12-17
is map_to_full_grid needed?


akhanf:
-this functionality might not be needed if we use (constrain_surf_to_bounding_box?)
-is this having any positive effect in any case? try without it..

jordandekraker:
closed with https://github.com/khanlab/hippunfold/pull/143

PR #142, opened 2021-12-22
mapfullgrid.sh no longer needed

jordandekraker:
As discussed in https://github.com/khanlab/hippunfold/issues/128, testing is working well without this script.

TODO: 
harmonize with https://github.com/khanlab/hippunfold/pull/130


akhanf:
Excellent news! I'll have a look through this too to make sure its a clean break, and remove any extra files/resources/dependencies we don't need as well. Then we can merge this in, and I can update #130 accordingly.

akhanf:
Actually, I think it will be easier to just drop mapfullgrid in #130..  I'll go ahead and do that..

jordandekraker:
aha yeah i had already started on trying to merge the two branches. Definitely easier to do on a new branch though, i'll test out https://github.com/khanlab/hippunfold/pull/143

PR #143, opened 2021-12-22
adds rules for composite warps from unfold and remove map to full grid

akhanf:
- moves all compose warps rules to separate rules file
- adds compose warp rules, and update to target rules
TODO:
 - [x] replace hacky orientation fix (this will take some effort to dig into -- if things work as is I suggest dealing with it later..)
 - [x] remove map to full grid (no longer needed #128)
 - [x] testing on real data


akhanf:
@jordandekraker this should be ready for testing now..

akhanf:
it still had some errors, but now it should be working.. I've confirmed that the unfold2native warps are correct now too (problem wasn't with orientation but just using the right nifti reference space when creating them)..

jordandekraker (approved):
Seems to be working on my end!

PR #123, opened 2021-12-05
workflow refactoring

akhanf:
Some substantial refactoring and maintenance changes to bring hippunfold up to date with snakebids, and to also adhere to snakemake workflow best practices. These could be considered breaking changes since they change the output file paths. Still a work-in-progress, but this is an overview of the changes:

* updates for compatibility with snakebids 0.4.0 (closes #121)
  * notably, changes output folder structure to prepend `results/` folder when in workflow mode (run with `snakemake`), and *not* prepend `results/` folder when in bids-app mode (run with `hippunfold` CLI)
  * in addition to this change, outputs are now placed in `hippunfold`and `work` folders, which are siblings. 
  * these changes mean that when run in workflow mode, all the outputs are in `results/hippunfold` and `results/work` (ie according to snakemake best practices to not crowd the snakemake root folder), and when run in bids app mode, all outputs are in `hippunfold` and `work`. 
* switches to poetry and pyproject.toml file for dependencies instead of setup.py (closes #122)
* updates to allow hippunfold to be run directly as a snakemake workflow 
* adds code quality and formatting checks
* adds release drafter workflow from snakebids (thanks @kaitj!)
  
- [x] run as snakemake workflow updates
  - [x] move workflow files to root folder instead of hippunfold module
  - [x] add config variables implicitly set by CLI as explicit config variables
  - [x] set these to defaults so dry-run can be run on test data
- [x] adds poetry for dependency management
  - [x] updates github workflows based on this
  - [ ] update install docs (`pip install` will still work, but for development will recommend using poetry instead of `pip install -e` which will not work)
- [ ] add quality checks, and fixes to pass the checks
  - [x] isort --check
  - [x] black --check
  - [x] flake8 
  - [x] pylint 
  - [x] snakefmt --check
  - [ ] snakemake --lint 
- [ ] add formatting github action 
  - [ ] black
  - [ ] snakefmt
- [ ] snakemake workflow fixes to pass `--lint`
  - [ ] add `log:` directive to every rule
  - [ ] add conda env for every `script:` rule 
  - [ ] once it passes, add to lint to quality and formatting actions
- [ ] find better solution for resolving paths to workflow files (using workflow.source_path was clunky)
  - [ ] current workaround uses `snakefile_dir` function just added, this could perhaps be a snakebids function
  - [ ] consider separating CLI args from directly configurable variables in different config yml files (may be snakebids changes)

PR #145, opened 2021-12-27
root dir change for snakebids 0.4.0

akhanf:
This PR is related to refactoring proposed in #123.

It changes output folder structure to prepend `results/` folder when in workflow mode (run with snakemake), and not prepend `results/` folder when in bids-app mode (run with hippunfold CLI)

In addition to this change, outputs are now placed in `hippunfold` and `work` folders, which are siblings (analogous to `results` and `work` from before).

These changes mean that when run in workflow mode, all the outputs are in `results/hippunfold` and `results/work` (ie according to snakemake best practices to not crowd the snakemake root folder), and when run in bids-app mode, all outputs are in `hippunfold` and `work`.

Note: more substantial refactoring could be done in later PR(s), where we can:
1. Add more subdirectories for organizing outputs (siblings to seg, surf, anat), like `coords` and `xfm`
2. Use the `temp()` directive throughout the workflow instead of using a separate work path. This would mean we could drop the `copy_to_results` rule, and instead of `--keep_work` can use the snakemake `--notemp` option to keep the temporary files


akhanf:
This PR now also changes the way we dry-run test the workflow, putting the tests in the github workflow instead of a separate python file and pytest. Did this because of some incompatibilities with the way we were testing and snakebids, but it should make adding new dry-run tests much easier..

jordandekraker (approved):

PR #146, opened 2022-01-04
Unfold DG

jordandekraker:
As per discussion in https://github.com/khanlab/hippunfold/pull/111, this branch will unfold the DG using laplace coords from templateshapeinjection step.

Current branch is working with `--density 32k`

TODO:
- [x] test on real data
- [x] add `unfold_template_autotopDG` den-7k, 2k, etc. files (currently only works for 32k)
- [x] consider renaming `{autotop}` to always be `label-{autotop}` for consistency
- [x] rename 32k -> unfoldiso; 7k -> 0p5mm; 2k -> 1mm; 400 -> 2mm


akhanf:
cool, I'll try going through this today too

akhanf:
Alright just had a run through on hcp 100307 -- I think the dentate midsurface is looking great! The inner and outer are quite wonky, but I think that is to be expected without doing any further smoothing or conditioning on the coords. 

I think for now we should actually just not generate the inner/outer surfaces for the dentate (and anything derived from them, e.g. thickness), and perhaps leave improving that for a future PR..  If you agree I can go through the workflow and make those changes. I have some other cosmetic BIDS naming changes I'd like to make too.

Some snapshots:

image

surface overlay without inner/outer
image

akhanf:
I can also go ahead and make the adaptive density surfaces too

jordandekraker:
Yup, that all sounds good to me!

akhanf:
ok I think this is just about good to go,  I am just running a final test on a single subject before merging in, but the test subjects I ran with the gen_unfold_template workflow looked great!

jordandekraker:
The changes look good to me! would you like me to give it a trial run as well?

Also, should we make this a new major release or should we wait for the other few PRs first? I'm eager to regenerate the plots used in the HippUnfold manuscript!

akhanf:
One minor thing j just noticed, we are still naming the .spec files with `hippunfold` as a suffix.. it's a bit non-informative, we could perhaps instead call this `surfaces` since it just contains all the surfaces

akhanf:
Yes I'm also eager to make a release! I've just changed the release drafter, so this next release will be v1.0.0 

I'll merge in #145 as well before the release since that is a breaking change too (since the output folder structure is changed slightly, `hippunfold` instead of `results`), then I can trigger the release workflow later today

If you want to do some final testing maybe it is best to do it from master after we have merged these two PRs just to make sure all is working before the release.

jordandekraker:
Great. I'll let you make these last few changes then so we're not both making the same commits. Once both PRs are merged i'll run some testing on my end! (will test the poetry install too)

PR #150, opened 2022-01-07
Docs from RST to Markdown

akhanf:
- changes docs to markdown instead of RST to make future editing/updating easier
- updates to readme, contributing etc..
- includes changes from #139 (can close that)

this is being rendered in:
https://hippunfold.readthedocs.io/en/markdown/


jordandekraker:
Now that we have a separate section for Contributors, we could probably simplify the Install instructions to just stick with singularity/docker (and maybe include a link to read more for contributors)

akhanf:
yes the contributors section is not meant to replace the installation section in the docs, that section is still where users should look, and the instructions there are up to date (they have been updated to remove `-e` from pip install, so will still work with our pyproject files).

akhanf:
ok i'll merge this in and can always make additional edits in another PR

PR #151, opened 2022-01-10
simple version for non-contributors

jordandekraker:


akhanf (approved):
Looks good to me, feel free to merge in. I've enable docs on this branch here: https://hippunfold.readthedocs.io/en/simplifiedinstall/index.html if you want to take a look before merging

jordandekraker:
For some reason these sections are not showing up in readthedocs: https://github.com/khanlab/hippunfold/tree/simplifiedinstall/docs/usage

akhanf:
ah good catch, looks like I broke those when converting to markdown -- I've reverted the two usage CLI doc files to RST to get it working again, let's see if that fixes it --

akhanf:
ok they're back, but I noticed there is a bug in grabbing the hippunfold CLI options, looks like it is something that broke with snakebids, I will need to look into that..

jordandekraker:
Nice yeah its back now, but the Command line interface page still looks funny - its showing the snakebids workflows instead of hippunfold options

akhanf:
ok actually was able to fix it by making a change to the `run.py` - so all good to go now I think

Issue #127, opened 2021-12-17
renaming/refactoring output directories


akhanf:
1. autotop directories in the work folder should live inside another folder
2. 'seg_{modality}' doesn't make sense - vol? (to go with surf?)
3. xfm or transform folder to store warps? (or warps?)
4. think about a coords folder (or subfolder?)
5. only allow user to specify 1 modality at a time  -- this would allow removal of the {modality} wildcard in filenames/folders and simplify outputs alot

6. workflow mode vs bids app:
```
workflow mode:
-results/sub-
-results/work
logs

bidsapp mode:
sub-
work
logs

config['root'] 
```

akhanf:
point 5. above is dealt with in #137

jordandekraker:
I think that we should change `seg` to `vol`, merge it with `anat` (which is typically nearly empty), and then place `warps` and `coords` subfolders inside that. I think this would be fine for both `work` and `root`. @akhanf thoughts?

Related bug: `from-unfold_to-corobl_mode-image_xfm` files are showing up in the results and i'm not sure why - I don't see them in the `common.smk` or in the `warps.smk`

akhanf:
instead of subfolders for warps and coords, we could just make them siblings, e.g. contents of `hippunfold/sub-001` could be:
```
vol
warps
coords
surf
```

actually regarding naming of the `vol` folder, I think I actually prefer to just call it `anat`, as it would contain the `_{T1w/T2w}.nii.gz` and `_dseg.nii.gz` images, so from a BIDS perspective that would be the required location for those BIDS derivatives files. It also leaves the door open for `func` and `dwi` folders, e.g. when we get around to adding rules to import and resample that data..

akhanf:
re: that `from-unfold_to-corobl_mode-image_xfm` file showing up in results, I can take a look, it could be using `root=root` in some generic rule.. 

In any case, we should probably finish merging in  #153 before we branch with more refactoring to make our lives easier

PR #153, opened 2022-01-11
workflow updates to allow T2w without T1w

akhanf:
Wanted to try hippunfold on some HSG data today (no T1w image), but there wasn't a nice way to do that currently, so I made these workflow updates to add rule for registration, then it actually wasn't that much more work to go all the way and make the native space generic to either T1w or T2w. Will test things out a bit more extensively, but the dry-run seems to be doing fine.. I think this will be useful for alot of users, @jordandekraker you probably have some data already you can test this on? We can also think about streamlining the CLI options a bit more if it works well (e.g. making reg_t2_to_template the default when modality==T2w..)

Changes:
- adds a reg_t2_to_template rule
- adds T2w CITI168 template image for this rule
- adds a --reg_t2_to_template CLI option to enable this rule (otherwise
default behaviour is to use T1w as usual)
- add a {native_modality} wildcard so rules that register or resample to
cropT1w/T1w space are now generic
- adds cropT2w to possible output_spaces
- in order to run T2w-only workflow, just need to specify:
`--reg_t2_to_template --output_spaces cropT2w`
- working with dry-run, and some initial testing on HSG data

TODO: 
- [x] test entire workflow on real data
- [x] make sure all dry-run testing works
- [x] add more dry-run test cases
- [x] update the documentation
- [x] revisit CLI names

### DIFF
--- a/.github/workflows/python-testing.yml
+++ b/.github/workflows/python-testing.yml
@@ -77,5 +77,8 @@ jobs:
     - name: Test cropseg bids, with path override, left hemi
       run: | 
         $HIPPUNFOLD - test_out participant -np --modality cropseg --path_cropseg test_data/data_cropseg_1hemi/sub-{subject}_hemi-{hemi}_dseg.nii.gz --hemi L
-
+    - name: Test T2w without T1w
+      run: |
+        $HIPPUNFOLD test_data/bids_singleT2w test_out participant -np --modality T2w --reg_t2_to_template --output_spaces cropT2w
+ 
 
--- a/.github/workflows/python-testing.yml
+++ b/.github/workflows/python-testing.yml
@@ -80,5 +80,8 @@ jobs:
     - name: Test T2w with T1w template registration
       run: |
         $HIPPUNFOLD test_data/bids_singleT2w test_out participant -np --modality T2w --t1_reg_template
- 
+    - name: Test T2w with additional T1w output space 
+      run: |
+        $HIPPUNFOLD test_data/bids_singleT2w test_out participant -np --modality T2w --output_space T1w
+
 
--- a/.github/workflows/python-testing.yml
+++ b/.github/workflows/python-testing.yml
@@ -77,8 +77,8 @@ jobs:
     - name: Test cropseg bids, with path override, left hemi
       run: | 
         $HIPPUNFOLD - test_out participant -np --modality cropseg --path_cropseg test_data/data_cropseg_1hemi/sub-{subject}_hemi-{hemi}_dseg.nii.gz --hemi L
-    - name: Test T2w without T1w
+    - name: Test T2w with T1w template registration
       run: |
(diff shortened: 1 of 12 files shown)


jordandekraker:
Good idea.

One more TODO: 

- [ ] update the 'No T1w images' section of the tutorials docs (https://github.com/khanlab/hippunfold/blob/master/docs/tutorials/standardBIDS.md#no-t1w-images)

I'll give this branch a test today

akhanf:
I'm a bit wary of adjusting the workflow automatically based on the available bids data for those same reasons, and the possibility that the workflow does something unintended by the user.. 

But I am thinking we could just make registration with T2 the default for {T2w,segT2w} workflows, e.g. instead of the `--reg_t2_to_template` option, we could have a `--template_reg_modality` option that can be used to override this to `T1w`.. 

Also, we could perhaps have the possible output_spaces options be {cropnative,corobl} instead {cropT1w,cropT2w,corobl}, and have the workflow set variables automatically so the user doesn't have to additionally change this option when selecting modality..

jordandekraker:
OK that makes sense to me! 

One other thought: I wonder if we should consider dropping the `--output_spaces` and simply leave it as `cropnative`?  `corobl` will always live in the `work` dir, and part of the reason I initially wanted it was just for when a cropped image was used as input (eg `--modality cropseg`) but I think that's already handled fine now. 
Just thinking about ways to simplify things for the user!

akhanf:
I see what you mean, but I think it's better to keep it in case we want to add more functionality, e.g. having options for bold or dwi..  It can be an option that the user doesn't need to worry about though, so we could perhaps leave it out of the higher-level documentation since the default would produce the desired result almost all of the time.. 

Another option I am wondering about is `--rigid_reg_template` -- this is basically required when we have reduced FOV data, as it fails badly if not.. I am wondering how it might affect things if we just make this option the default? (e.g. replace it with a `--use_affine_reg_to_template` that is disabled by default)?

jordandekraker:
OK sounds good. After this PR I'll make a few updates to the example Tutorials. 

I think we want affine to be the default for registering to the CITI template, but I could be convinced that rigid is better

akhanf:
I agree affine is probably preferred for whole brain data .. maybe we can just call the `--rigid_reg_template` option something like `--reduced_fov` to make it clear this must be used when using reduced FOV data?

akhanf:
ok I made some CLI changes, let me know what you think..

jordandekraker:
Yeah this looks great! I'll make some updates to the tutorials and commit them before we merge

jordandekraker:
OK i made some basic updates. The docs will still need a few updates to include things like images of DG surfaces and changes to filenames, as in https://github.com/khanlab/hippunfold/issues/127, but I think this should cover everything in the current PR

akhanf:
We can continue to make minor/patch releases afterwards (for docs, minor updates etc), but should try to have the major breaking changes in this v1.0.0 release

jordandekraker:
Agreed!

PR #155, opened 2022-01-12
output refactoring, removed datatype=seg folders, added coords, warps

### LINKED ISSUE #127: renaming/refactoring output directories

akhanf:
closes #127 

To do:
 - [x] verify final output paths with a test run on real data
 - [x] update docs
 - [x] fix the specific presence of file (in results, but should be work?) mentioned in #127


akhanf:
@jordandekraker let me know what you think, I haven't verified everything myself yet --

akhanf:
yes was thinking the exact same thing re: surf.. I wonder what we should keep default, 0p5mm or 1mm? 
 actually I already have the second point committed looks like I just forgot to push

akhanf:
oops nevermind re: the last point, it was already committed

jordandekraker:
I think most people these days will be working with data <=0.8mm for anatomical work, so I vote we keep `0p5mm` as the default (plus it looks nicer). Anybody doing fMRI will probably want to use `2mm` anyway

akhanf:
> 

oops I am getting confused with my PRs -- this additional edit wouldn't belong here anyhow, so lets just merge this in as is..

> Re: the output-spaces changes I made I'd like to alter it slightly to be: default: output-spaces=['native'] (the behaviour currently, e.g. T1w or T2w depending on what modality is used)
> 
> and if you do: output-spaces=['native','T1w'] , (e.g. --output-spaces native T1w ) it will additionally output the data in T1w space,
> 
> and if you do: output-spaces=['T1w'], (e.g. --output-spaces T1w) it will **only** output data in T1w space..
> 
> This would allow us to still use T2w to segment, but get everything in T1w space.. I'm going to try my hand at this as a final tweak to this PR

jordandekraker:
OK i'll make a few quick updates to the readthedocs. I'll still need to make some new images to show the DG surface too, but I'll make that a separate issue

akhanf:
actually I ended up making the output-space change here too, was just easier to do it here instead of another branch, along with a couple other minor refactoring changes, most notably, removing desc-cropped from corobl files

jordandekraker:
Just made some changes to the readthedocs, let me know what you think. I didn't see your last commit so I'll have to see if the output naming needs updating again

jordandekraker:
This is looking good to me! do we need the `surf/dscalar.nii` files?

akhanf:
yeah those cifti files are more useful sometimes than the gifti ones, as can do cifti-math with them.. I know it's a bit redundant to have both the gifti metrics and cifti scalars for the same thing, but I think each are probably useful in their own ways.. I'll take a quick look at the docs too and merge it in..

PR #159, opened 2022-01-13
allowing qc to use corobl

jordandekraker:
Noticed that the qc folder was nearly empty when the output space is corobl (eg with `--modality cropseg`). This should fix that (slice views will have no background image)


akhanf (approved):
Looks good to me - though the name of the wildcard isn't actually important (ie technically it just has to be consistent between inputs and outputs of a rule) but good to have some consistency across rules or for semantic purposes..

jordandekraker:
was still throwing some errors with different modalities, but i think i have it fixed now. I think the current Check error is maybe a linting thing? I'll check it out

akhanf:
taking a look at this now -- after fixing the formatting errors the hippb500 workflow was broken, so will adjust accordingly..

PR #160, opened 2022-01-15
added logging

akhanf:
- this is a cosmetic fix (doesn't change outputs, except for log files)
- the logs printing were screen were an eyesore for some time, so went through and tried to add logs to any rules that were doing this


jordandekraker (approved):
Ahh this should make things a lot neater! I didn't test yet but it all looks good to me

akhanf:
I ran a test on real data to confirm earlier so should be good

PR #172, opened 2022-02-18
myelin mapping (T1w/T2w)

akhanf:
Here is a first crack at incoporating myelin mapping outputs in hippunfold.

Adds a `--generate-myelin-maps` option that enables it. Can be used with `--modality` set to `T1w` or `T2w`, uses images in the corobl space to calculate the T1w over T2w, then maps that to surface using ribbon-style mapping (for both hipp and dentate).

I made the spec file rule generic to hipp vs dentate to incorporate this as well, adding some config variables for `cifti_types` and `gifti_types`. 

To do:
 - [x] testing on modality other than T1w 

Questions (@jordandekraker, @royhaast ):
 - [x] is ribbon the best approach here? 
 - [x] any additional pre/post-processing we should do on the T1w/T2w maps?
 - [x] is it worthwhile to add a way to import existing T1w/T2w ratio maps?

note: this PR also includes a minor update of dependencies, as older version of snakefmt was doing funny things..


jordandekraker:
Sorry for the slow response, but this is working well on my tests!
I don't think we need any additional processing. As for importing an existing ratio, i think that should probably be the same as things like a T1map, FA, MD, fMRI, etc.
I don't think there's too much to add to the Docs here either since the CLI will be updated automatically

PR #184, opened 2022-05-27
expose crop_native_box as a config and CLI variable

akhanf:
was hard-coded to 256x256x256vox


Bradley-Karat:
Test case worked and the spacing issue is now fixed in the YAML

akhanf:
awesome thanks, will merge this in now..

PR #186, opened 2022-05-30
Freesurfer atlas

akhanf:
Adds freesurfer-based subfield unfolded atlas, using this workflow to create the new files: https://github.com/khanlab/hippunfold-create-atlas/tree/freesurfer  (ie, mapping automated labels at the midsurface to the unfolded atlas, and doing maxprob over all subjects)

- [ ] add to docs
- [ ] consider removing labels (?) -- I kept all but amygdala (labels 7001+) from the freesurfer subfields LUT, but we could potentially remove all that we don't want to get mapped into the CA/Sub/DG ribbon..


akhanf:
Actually it looks like FS spits out additional sets of segmentations: 
https://surfer.nmr.mgh.harvard.edu/fswiki/HippocampalSubfieldsAndNucleiOfAmygdala
and the "CA" set is probably the most suitable here, as it already folds in the SRLM labels. I'll regenerate a new label nifti with those.

jordandekraker:
Sounds good! I would still maybe remove the `fissure` label since in our protocol we include it as part of SRLM. 

Also my test case seems to have worked perfectly! I'll try again with the new labels when its committed

akhanf:
Looks like using the "CA" set of labels dealt with things well, the only labels that show up are presub,sub,CA1,CA3,CA4.  There is a tiny island of parasubiculum (and islands of CA1, Sub) in the most distal end, but I think it's fine to just leave that as is since easier to explain. I'll commit these in a sec, but here are screenshots fyi

image
image

PR #194, opened 2022-06-07
Drop generic copy_to_results rule

akhanf:
Drops the copy_to_results rule and adjusts workflow accordingly..

Note that this may end up with more files in hippunfold dir (e.g. corobl space surfaces)


akhanf:
@jordandekraker a heads up, in case you have any other active branches, as this makes a number of refactoring changes while replacing the copy_to_results rule..

jordandekraker (approved):

PR #200, opened 2022-07-11
allows output of subfields using multiple unfolded atlases

jordandekraker:


jordandekraker:
I thought this would be a small fix and I could do it via browser. It ended up taking a full test environment and cases. 

the gifti .spec files didn't have an {atlas} wildcard, and so i had to add a function that would tack them on.

PR #201, opened 2022-07-26
added missing canonical surfaces for label-dentate

jordandekraker:
Generated these from BigBrain (hemi-R) for the purpose of easy and generic visualization of surfaces and surface data


akhanf:
Cool. We should think about adding these to the main workflow too with subject data mapped to them (as optional or default), and where we might put them in the file/folder structure.

jordandekraker:
So far i've found they're most useful for visualizing second-level, cross subject analyses. 
I'm going to try and keep this up-to-date to reflect that:
https://github.com/jordandekraker/hippunfold_toolbox/blob/main/Python/tutorial_micapipe.ipynb

akhanf:
@jordandekraker I think this PR is replaced by #203, so feel free to delete this if so

Issue #202, opened 2022-07-27
unfolded DG template not rectangular


jordandekraker:
https://github.com/khanlab/hippunfold/blob/master/hippunfold/resources/unfold_template_dentate/tpl-avg_space-unfold_den-2mm_midthickness.surf.gii

Its more of a trapezoid really. This only really matters for aesthetic purposes (eg. generating folded/unfolded surfaces).

Should edit this script to ensure the corner vertices are never removed:
https://github.com/khanlab/hippunfold/blob/master/gen_unfold_template/gen_adaptivesurfs.py

PR #203, opened 2022-07-27
Unfolddentatetemplate patch

jordandekraker:
should be a quick fix for https://github.com/khanlab/hippunfold/issues/202


akhanf:
You may have to walk me through this first. Also how does this relate to #201 ? (does it replace that PR?)

jordandekraker:
The Matlab code finds the corner vertices and moves them to the outer bounds of the surface (ensuring that the corner vertices make a rectangle instead of a trapezoid). I accidentally made this branch off of #201 so yes it also includes that change! 
Can go over it briefly in our weekly meeting too.

jordandekraker:
Re Friday meeting: instead of moving the corner vertices to make a rectangle, we could add another vertex there (changing the number of `den-2mm_label-dentate` vertices). Personally I don't think its worth it to change, but I wanted to make a note of it here so we don't forget

akhanf:
Yeah I agree now, isn't worth changing the surface to add a vertex as would break compatibility, and it's mostly for esthetic purposes (interpolation shouldn't be affected much)

akhanf (approved):
looks good to me -- as I mentioned on Friday, these scripts and surfaces could just as easily live in hippunfold-toolbox instead, but fine to merge in for now

PR #208, opened 2022-08-13
bugfix: subfield+tissue for new atlases, and tsv files

akhanf:
Two related fixes:

1. New atlases were still using the bigbrain dseg.tsv for generating the
subfield volumes.tsv files. These have now been added.

2. The way the postproc tissue labels (dg, srlm, cyst) were being added
to the subfields was also hardcoded for the bigbrain atlas. Have now
added a section in the config file to customize what labels these map
to, so that e.g. the dentate gyrus is merged into the existing label
number, and that the SRLM/Cyst labels do not conflict with the labels,
and are maximal (so they override the others). Previously the magdeburg
subiculum label (8) was same as cyst, so these became merged.

### DIFF
--- a/hippunfold/config/snakebids.yml
+++ b/hippunfold/config/snakebids.yml
@@ -345,6 +345,28 @@ atlas_files:
     label_list: resources/freesurfer/freesurfer_labellist.txt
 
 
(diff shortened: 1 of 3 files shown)


akhanf:
tested and working, so merging this in

PR #211, opened 2022-09-08
bugfix: subfields label.gii wasn't being generated properly

akhanf:
bug introduced in a recent PR, had to convert the metric gii to a label gii after volume-to-surface-mapping


akhanf:
also adds subfields dlabel file back into spec files (was not being generated before, also from a recent PR)

PR #216, opened 2022-09-13
Added missing container for hippdwi_to_template

### LINKED ISSUE #215: resample_hippdwi_to_template container
Was debugging with @Bradley-Karat and Mahdieh. Was encountering an error for missing `c3d` command. Noticed that the `resample_hippdwi_to_template` rule doesn't specify a container. Not sure if it needs to be specified?

kaitj:
This PR adds the `autotop` container to fix the missing command error for `c3d` when the `hippdwi_to_template` rule is run.

Resolves #215

### DIFF
--- a/hippunfold/workflow/rules/preproc_hippb500.smk
+++ b/hippunfold/workflow/rules/preproc_hippb500.smk
@@ -21,6 +21,8 @@ rule resample_hippdwi_to_template:
             suffix="b500.nii.gz",
             **config["subj_wildcards"]
         ),
+    container:
+        config["singularity"]["autotop"]
     group:
         "subj"
     shell:


akhanf (approved):

akhanf:
actually wait, now having some strange issues with the version of c3d in that container - the rule runs fine with c3d installed on cbs server.. Will see if maybe upgrading to autotop 0.4.2 fixes it

akhanf:
nevermind, the issue I was just having is actually a separate one, related to reading in files from home directory (or `~/graham`) when using a container rule in snakemake -- it changes the home directory when running singularity, so need to have `SINGULARITY_BINDPATH` that includes `/home/ROBARTS:/home/ROBARTS`.  I'm in the process of making this the default on the system, but will make sure we note this in the documentation too..

So I'll merge this in!

PR #220, opened 2022-11-09
aligns laplace_coords.py behaviour to laplace_coords_withinit.py

jordandekraker:
This was causing a bug where extra-hippocampal areas contained Laplace coordinates, which was messing up the mask generation in create_warps.py. This should fix that by setting all background labels back to NaN, and this is now in line with the behaviour of laplace_coords_withinit.py.


jordandekraker:
Tacking on an additional related issue into this PR: 
When using `--modality cropseg`, sometimes the border of the hippocampus reaches the border of the crop, and the Laplace solver's nan_convolve fills a value of 0.0. Switching this to a fill value of NaN instead produces the desired behaviour (ie. ignoring values outside the cropped area).

akhanf (approved):

akhanf:
@jordandekraker ok to merge this in?

jordandekraker:
tested and working on my end!

PR #221, opened 2022-11-10
Download rule in the workflow

akhanf:
- uses snakemake's http remote to download files from zenodo url
- replaces the `hippunfold_download_models` script
- location of files remains the same for now
- reduces the docker container size from 12.84GB to 3.37GB  - can try it out in docker://khanlab/hippunfold:download-workflow

TODO:
- [x] dockerfile updated
- [ ] maybe add a new CLI argument --model-dir
- [ ] come up with solution for dealing with nodes without internet access (maybe just a conditional to check connection, and skip the rule otherwise)
- [ ] think about changing default location of models - is resources folder a better default than home directory? more likely to work, but space-inefficient if the user never overrides.. Or maybe for reproducibility it is worthwhile having the models appear with the results?


akhanf:
I'll merge this in later, want to test things a bit more and not have to revamp the docs before the resubmission..

akhanf:
quick update -- for whatever reason (have noticed it elsewhere too) the dropbox urls don't work with snakemake http remote... they end up downloading an ~800K file with html info in it, but not the file itself.. This same code was working back in November so not sure what changed..

akhanf:
I'm going to move introduction of the synthseg model to another PR, as this one is stalled for now..

PR #224, opened 2022-11-17
defined in mm instead of vox

jordandekraker:
this should be more robust to different input image resolutions. 52mm iso should be enough to cover the whole hippocampus.

### DIFF
--- a/hippunfold/config/snakebids.yml
+++ b/hippunfold/config/snakebids.yml
@@ -190,9 +190,15 @@ parse_args:
     default: false
     action: store_true
 
+  --crop_native_res:
+    help: 'Sets the bounding box resolution for the crop native (e.g. cropT1w space). Under the hood, hippUnfold operates at higher resolution than the native image, so this tries to preserve some of that detail. (default: %(default)s)'
+    default: '0.2x0.2x0.2mm'
+
   --crop_native_box:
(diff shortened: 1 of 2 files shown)


akhanf:
Isn't this a much smaller bounding box? (256 x 0.3mm = 76.8mm)

jordandekraker:
I believe voxel size was relative to the input image, so in my case I had a 0.8mm being increased 400% to 0.2mm isotropic in the cropT1w, which worked out to 51.2mm. Just testing now so I'll confirm whether this is a good new default or not

akhanf:
Ah yes forgot about that, yes you're right!

jordandekraker:
while we're at it, i wonder if its worth setting a fixed voxel size, or maybe just making it relative to the `space-corobl`?

akhanf:
I'm good with changing it to be a fixed size (rather than 400% relative), specified in a config file and/or cmd line arg?

jordandekraker:
Great, I'll set that up right now! I'll keep the default to 0.2mm (or 0.3mm?) to minimize loss when rotating the corobl back to native (which is usually about 25-30deg sagittal rotation)

akhanf:
yep I'm good with either one

akhanf (approved):

PR #228, opened 2023-02-13
Unfold reg

jordandekraker:
Performs unfolded registration (ants SyN) in 2D unfolded space. only available for the new atlas `histologyReference2023`, which is now the default behaviour. 

Unfolded registration is based on atlas thickness, curvature, and gyrification. These now need to be calculated twice per subject (before and after alignment). All surface meshes have the alignment applied, such that vertices are all more closely matched (or homologous) between subjects. 

See WIP manuscript draft


jordandekraker:
Just remembered something else TODO:

- [ ] concatenate unfolded space transform onto native-to-unfolded transform (probably in python since formats are different)

akhanf:
Didn't you use ANTS to perform registration? Should be the same format if so..
I haven't had a chance yet to try out this branch but will put it on my todo list. Was thinking it would be good to have some benchmark datasets we use for larger-scale testing, e.g. to see how major changes might be affecting subfield volumes, thicknesses etc.. I can make a separate workflow (maybe in another repo) for automating that testing..

akhanf:
Ah ok. Yes extending it to 3D would be something custom probably easier written in numpy, but once you have that 3D reg in the unfolded volume space, you could add it to the chain of concatenation as we do already..

jordandekraker:
OK the warps are now working for all test cases, and i can see that applying them makes a small but noticeable difference to unfolded T2w data.
Not sure what this `snakfmt` quality check error is though

PR #230, opened 2023-03-05
fix bug in crop native box

akhanf:
This was being interpreted as 52x52x52vox instead of 52x52x52mm (could be a c3d bug or limitation in the `-pad-to` command. So had to change this back to voxel definition, 256x256x256vox. This is ok since at least the voxel resolution can be specified separately in mm.


akhanf:
closes #231

PR #237, opened 2023-03-31
Fix quality action

kaitj:
@akhanf mentioned that the quality action has been failing when trying to install a package (see [here](https://github.com/khanlab/hippunfold/actions/runs/4546842822)).

Turns out this is due to a change made in the latest version of _Poetry_ (the actions workflow doesn't pin the Poetry version used). As of `v1.4.1`, packages are required to use modern installation (see: https://github.com/python-poetry/poetry/issues/7686). This is potentially problematic when relying on older versions of packages and for packages who are still using non-modern installers. This PR fixes this by adding an additional step to disable the modern-installation requirement in the Poetry config.

Also used this PR as an opportunity to: 
* update the versions of action steps that was still using older versions of node.js
* Add a manual trigger for the python-testing workflow

### DIFF
--- a/.github/workflows/bump_version.yml
+++ b/.github/workflows/bump_version.yml
@@ -7,7 +7,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@master
 
       - name: Update changelog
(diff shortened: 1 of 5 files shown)


akhanf (approved):
Thanks Jason!

PR #240, opened 2023-06-08
Warps fix

jordandekraker:
OK this is a small fix and results in surfaces that are shifted slightly towards the middle, but I think its still very important.

I started getting this issue in some histology unfolding, and couldn't fix it by modifying the segmentations or the laplace coordinates:
innersurf_displaced

This displaced surface didn't even line up with any of the x-y-z-axes.

Eventually I realized it went as far back as the inner-most layer of the warp:
warpfield_displaced

As it turns out, some of this layer was outside of the convex hull spanned by the Laplace coordinates. I played around with the interpolation/extrapolation methods, but no luck. Finally i realized that the extrapolated areas that were working just fine, and the error was occuring because there were a few Laplace coordinates present in the inner-most unfolded layer. These values must have been really similar to the values in the layer above them despite being separated by a whole IO layer, and so the linear interpolation was being thrown way outside. 

This fix should avoid the issue altogether, and has worked in a few similar histology samples now. I think it may have also caused an issue in some previously processed MRI data as well, so I'm very happy to have it solved!

### DIFF
--- a/hippunfold/workflow/scripts/create_warps.py
+++ b/hippunfold/workflow/scripts/create_warps.py
@@ -119,8 +119,8 @@ points = (
 # get unfolded grid (from 0 to 1, not world coords), using meshgrid:
 #  note: indexing='ij' to swap the ordering of x and y
 (unfold_gx, unfold_gy, unfold_gz) = np.meshgrid(
-    np.linspace(0, 1, unfold_dims[0]+2)[1:-1],
-    np.linspace(0, 1, unfold_dims[1]+2)[1:-1],
+    np.linspace(0, 1, unfold_dims[0]),
+    np.linspace(0, 1, unfold_dims[1]),
     np.linspace(0, 1, unfold_dims[2]+2)[1:-1],
     indexing="ij",
 )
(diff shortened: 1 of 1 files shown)


akhanf (commented):
Hi Jordan,

I guess this is effectively is  adding a little fudge-factor so instead of being from 0 to 1, goes from 0+epsilon to 1-epsilon..  That sounds fine to me, but just the way it is parameterized (i.e. by linspace over more points and dropping the ends), rather that simply adding the epsilons, makes it dependent on how large the dimension is, and also makes the code less clear. For your cases, are you using unfolded grid of 256x128x16? If so then the epsilon would be ~0.05.

So we could perhaps define this config variable:
```
unfold_crop_epsilon_fractions:
 - 0
 - 0
 - 0.05
```
then adjust the code to be something like:
```
np.linspace(0+snakemake.config['unfold_crop_epsilon_fractions'][0],
            1-snakemake.config['unfold_crop_epsilon_fractions'][0], 
            unfold_dims[0]),
np.linspace(0+snakemake.config['unfold_crop_epsilon_fractions'][1],
            1-snakemake.config['unfold_crop_epsilon_fractions'][1], 
            unfold_dims[1]),
np.linspace(0+snakemake.config['unfold_crop_epsilon_fractions'][2],
            1-snakemake.config['unfold_crop_epsilon_fractions'][2], 
            unfold_dims[2]),
```

jordandekraker:
Yup, that's a good description of what this is doing, and I'll go ahead and implement your suggested code change!

jordandekraker:
test case now working, probably ready to merge

akhanf (approved):
yes looks good to me! feel free to do a squash merge to make 1 clean commit

PR #243, opened 2023-07-04
Synthseg model

jordandekraker:


jordandekraker:
will merge into https://github.com/khanlab/hippunfold/pull/228 instead

PR #244, opened 2023-07-04
Synthseg model

jordandekraker:
This looks like a nice update that adds an optional (experimental) model.

I think the change to the L/R flipping method is good for debugging the `work/` dir, but I don't think it will affect any of the final outputs since they get flipped twice anyway. Can you confirm @akhanf?

Only other thing I might suggest is naming the model synthseg_v0 or something like since we may update it later on (I think I have a slightly more up-to-date model that I could sub in too)


jordandekraker:
The new unflipping method was generating some problems, so made fewer steps utilize the hemi-Lflip space. Namely, only shape_inject.smk and nnunet.smk used flipped space (autotop.smk no longer does).

PR #247, opened 2023-07-24
Download models inside workflow

akhanf:
This replaces PR #221, and adds functionality to download models within the workflow. 
Currently a WIP as doing some testing on it now..

### DIFF
--- a/.github/workflows/push_container.yml
+++ b/.github/workflows/push_container.yml
@@ -43,7 +43,6 @@ jobs:
             ghcr.io/${{ github.repository }}
           flavor: |
             latest=auto
-            suffix=_synthseg
       
       - name: Build and push Docker images
         uses: docker/build-push-action@ad44023a93711e3deb337508980b4b5e9bcdc5dc
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,11 +4,11 @@ MAINTAINER alik@robarts.ca
 
 COPY . /src/
 
-#pre-download the models here:
-ENV HIPPUNFOLD_CACHE_DIR=/opt/hippunfold_cache
+# avoid pre-downloading the models to make for lighter container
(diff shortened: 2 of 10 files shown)


akhanf:
@jordandekraker looks like the the snakemake http remote doesn't like the dropbox URL for the synthseg models, I'll move them to zenodo (better for consistency and longevity anyhow)..

akhanf:
OK looks like this is working now, have tested it on graham too. 
Couple things though: 

1. As before, you need to set HIPPUNFOLD_CACHE_DIR to the folder where you want the models. However, now, if you are running from a hippunfold container, you would need to make sure that env var is set inside the container too, by setting either: `SINGULARITYENV_HIPPUNFOLD_CACHE_DIR` or `APPTAINERENV_HIPPUNFOLD_CACHE_DIR`. I think Apptainer is backwards compatible with the use of SINGULARITYENV but it may ultimately be deprecated..  I believe compute canada has migrated completely to apptainer anyways?

2. If a downloaded model timestamp is earlier than the timestamp on the remote file on zenodo, then it will be re-downloaded.. 

3. I haven't been able to confirm how this works on a node without internet (the graham node I ended up testing on seemed to have internet connectivity).. so I think it may be worth testing it a bit more..

4. It wouldn't be hard to make some versions of the container that have a single model, e.g. for the most popular ones like T1w, T2w, and synthseg.. That could make life easier for people as they wouldn't need to manage a HIPPUNFOLD_CACHE_DIR..

jordandekraker:
That all sounds good to me, but maybe we should make a note in the documentation (maybe the FAQ) about the default `HIPPUNFOLD_CACHE_DIR` being in the `~/.cache` directory. I don't think its worth building multiple containers with the other models though.

I was also thinking we should try to build this soon since a few people might `singularity pull hippunfold:latest` hoping to get v1.3.0. 

I won't be around much for a few weeks though, sorry, on vacation!

Issue #239, opened 2023-05-16
small bug present when visualizing with freeview


akhanf:
From a user:

When we want to apply the file "desc-subfiels_freeview_dseg"
on the segmented image, SRLM and the dentate gyrus do not appear. By
modifying the desc-subfiels_freeview_dseg.tsv file and replacing the
spaces with "_", it respects the freeview nomenclature and solves the
problem.

PR #256, opened 2023-08-11
rename histologyReference2023 to multihist7

### LINKED ISSUE #254: Naming of histologyReference2023
@jordandekraker I didn't pay attention to this when reviewing the initial PR, but the naming for this atlas could be improved. It seems nit-picky but it's something we would have to live with for some time, so I think it's not too late to finalize it (ie before paper is published, and before many people start using it) 

- the use of the word "Reference" is redundant (any atlas is being used as a reference)..
- the name is really long, and uses camelCase which goes against the existing naming scheme and bids style
- we haven't been using years in any naming for other atlases etc so the 2023 feels out of place..
- what is quite novel over the previous `bigbrain` atlas is this is built from 7 datasets instead of the bigbrain pair of hippocampi, but this isn't apparent from the name

What do you think of renaming this atlas to be `histology7` instead? It is shorter, and conveys more meaning than the previous name.. or `multihist7`? 

By the way, what about also publishing the probabilistic labels for each subfield?

akhanf:
Closes #254.

Shorter and more meaningful name.

### DIFF
--- a/docs/getting_started/docker.md
+++ b/docs/getting_started/docker.md
@@ -10,20 +10,20 @@ Open your Windows Command Prompt by clicking the bottom left `Windows` button an
 
 Pull the container (this will take some time and storage space, but like an installation it only needs to be done once and can then be run on many datasets):
 
-    docker pull khanlab/hippunfold:v1.2.0
+    docker pull khanlab/hippunfold:v1.3.3
 
 Run HippUnfold without any arguments to print the short help:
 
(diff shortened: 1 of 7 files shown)


jordandekraker:
This should also close https://github.com/khanlab/hippunfold/issues/239

I'm including it in this PR since its a similar problem (ie naming)

jordandekraker:
This solved my Freeview issue, let me know if you're happy @akhanf

PR #262, opened 2023-09-22
Change in dependencies

akhanf:
- This PR makes use of an updated autotop_deps container (now called hippunfold_deps) that includes the python script dependencies too
- This allows us to then remove those dependencies from this package, making it much lighter and easier to install - ie you can run hippunfold if you just have snakemake+snakebids installed, you don't need all the extra dependencies (nnunet, astropy etc..)
- Doesn't really change anything from the user end, as containers have been required for running hippunfold anyways, this just ensures that *all* the rules use containers..


akhanf:
Note: we could put hippunfold on pypi eventually too, but will need to slim it down by putting some resources on zenodo or elsewhere (the package with templates etc was too big for pypi last I checked, and it has only grown)..

akhanf:
end-to-end tests worked fine for me with this, so going to merge it in..

jordandekraker:
was going to mention, I also tested this on my end and it worked perfectly

PR #269, opened 2023-10-31
new branch for cleaner PR

jordandekraker:


jordandekraker:
@Bradley-Karat we can put our in-progress / updated segmentations here now: https://osf.io/v8acf/
Its possible that editing these files will create a new hash in the config (eg [here](https://github.com/khanlab/hippunfold/blob/5974a5f5e1b5861a8e4c4205efae4423a5ad5a6e/hippunfold/config/snakebids.yml#L474)). If so, I'll help you work out the new hash since it involves multiple steps

jordandekraker:
TODO: reformat downloads, etc to accept multiple `atlas` arguments

PR #264, opened 2023-10-02
Marmunfold

akhanf:
## Marmoset hippocampal unfolding

- Adds a new template-based segmentation (alternative to nnunet) option, `--use-template-seg`
- Adds two new templates, MBMv2 (from ex vivo single hemisphere data), and MBMv3 (from in vivo group template) (using images from here: https://marmosetbrainmapping.org/atlas.html)

e.g. these options can be used to segment/unfold a marmoset in vivo brain MRI: ``` --template MBMv3 --use-template-seg --crop-native-res 0.05x0.05x0.05mm  ```

The manual tissue dsegs were created by manual correction and a number of additional registration/image processing steps (thanks @Bradley-Karat @jordandekraker for the initial MBMv2 segmentation in the native dti space) 

This also adds a simple snakefile for producing the coords for the templates from these dsegs (since those are required for initialization), was done manually in Matlab before I believe.. 

Note 1: Haven't tested this extensively on new (individual marmoset) datasets yet, but I am also creating a repo (`marmunfold`?) with hippunfold results run on the templates themselves, for use in group studies (eg that already use the MBMv3 space), specifically for an Everling lab collaboration

Note 2: the tissue dsegs could use a little bit more tweaking, there is a discontinuity in the head that needs to be resolved. @Bradley-Karat or @jordandekraker would be great if you are able to have a look, I might run out of time to get to that..

PR #270, opened 2023-11-16
Add testing for other python versions

kaitj:
The current testing workflow only supports Python 3.8 - this adds additional hippunfold-supported Python versions (3.7 and 3.9) for testing. Given it's a dry-run test, it may not catch any run-time issues, but would potentially catch dependency issues at the installation step.

Additionally, a few minor maintenance things:
* Versioned action steps instead of the using the `main` branch (which may be a dev branch)
* Where possible, switched to reusable workflows for easier maintenance and reduce the number of lines in the file.

### DIFF
--- a/.github/workflows/bump_version.yml
+++ b/.github/workflows/bump_version.yml
@@ -1,13 +1,16 @@
+---
 name: Bump version
 on:
   push:
     branches:
       - master
+
 jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@master
+      - name: Checkout main branch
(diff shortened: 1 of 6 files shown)


jordandekraker:
Amazing, thanks Jason!

Could this be related to the https://github.com/khanlab/hippunfold/pull/264 check failing?

kaitj:
Hmm, that looks like its an issue with building the documentation. Not entirely sure why that is happening given these other PRs are still building the documentation without issue. I'll take a closer look and see if I can get that going again.

jordandekraker (approved):

PR #284, opened 2024-03-05
Docpatch

jordandekraker:


kaitj:
_(From message in commit 565n669)_

RTD was still using py3.7, which on its own probably would have been okay. However, `pybids/bids-validator` now has a minimum python requirement of py3.8, which I suspect made some changes to `functools.lru_cache`. These changes caused RTD to error out when trying to read the `bids-validator` code.

PR #300, opened 2024-06-12
vertex cleaning, especially for DG

jordandekraker:
Here I tried to clean up some misplaced vertices, which occurs especially often in DG. to do this, I employ s strategies:

1) in addition to filling NaN vertices, we also find bad vertices, which are defined by their distance to a smoothed surface
2) create_warp.py now uses a 256,128,16 laplace coord grid, instead of 0-1x0-1x0-1. This should make the determination of  neighbours a bit cleaner

NOTE I also played around with some config options. I'm not sure whether this helps, so i'm going to undo the next changes in my next commit. We should test on more cases to find out.


akhanf:
With PRs like this that impact all workflows (ie not adding a new feature, but changing how a existing computation is done), it would be good to set-up some regression testing (wet-run, not just dry-run) to assess how much an impact these changes make. 

Would be good to do this for a variety of datasets too.  Would have to decide on a few things (what the tests will actually evaluate, what datasets, how to perform the tests efficiently on github actions), so I think a separate issue needs to be spawned.

jordandekraker:
agreed. I'm doing some local testing, but it would be good to have something more systematic.

Similarly, i'm adding a bigbrain `template` (at 40um which is what we did in the BigBrain neuroimage paper) to the OSF repo so i can see how dentate processing works there. Will PR it into this branch shortly (or maybe not so shortly, it takes a long time to get the Laplace solutions at that resolution!). We could add that as a possible wet-run as well, but its a bit niche and heavy so maybe i'll stick to just testing it locally.

jordandekraker:
closed with https://github.com/khanlab/hippunfold/pull/306

PR #305, opened 2024-06-26
Naturalinterpolation

jordandekraker:
This lovely interpolation function is fast and initial tests show it working extremely well, solving all issues in https://github.com/khanlab/hippunfold/pull/300 without the need for any postprocessing of surfaces. I tried many things to clean these surfaces, including different interpolation functions, but I believe this fix now finally solves the root cause. 

This is a breaking change since outputs will now be slightly different (and better quality). I see no downsides or anticipated conflicts. 

This is also now in-line with the original Matlab code that was used for interpolation ([hippo_autotop](https://github.com/jordandekraker/Hippocampal_AutoTop))

Still doing a wet-run test.


jordandekraker:
will move this to a new PR without commits from #300

PR #306, opened 2024-06-26
use naturalneighbour interpolation instead of linear barycentric

jordandekraker:
This PR adds only the best changes from https://github.com/khanlab/hippunfold/pull/300 and https://github.com/khanlab/hippunfold/pull/305. 

Notably, from https://github.com/khanlab/hippunfold/pull/300 no changes are kept except those that make surface config easier to use. Also interpolation is performed in 256x128x16 unfolded space instead of 0-1x0-1x0-1 unfolded space. Other changes are no longer needed since they are handled in https://github.com/khanlab/hippunfold/pull/305.

From https://github.com/khanlab/hippunfold/pull/305, here are the major changes:

Naturalneighbor interpolation is in create_warps.py. This lovely interpolation function is fast and initial tests show it working extremely well, solving all issues in https://github.com/khanlab/hippunfold/pull/300 without the need for any postprocessing of surfaces. I tried many things to clean these surfaces, including different interpolation functions, but I believe this fix now finally solves the root cause.

This is a breaking change since outputs will now be slightly different (and better quality). I see no downsides or anticipated conflicts.

This is also now in-line with the original Matlab code that was used for interpolation ([hippo_autotop](https://github.com/jordandekraker/Hippocampal_AutoTop))

Still doing a wet-run test.

### DIFF
--- a/hippunfold/workflow/rules/common.smk
+++ b/hippunfold/workflow/rules/common.smk
@@ -24,23 +24,40 @@ def get_modality_suffix(modality):
 
 
 def get_final_spec():
-    specs = expand(
-        bids(
-            root=root,
-            datatype="surf",
-            den="{density}",
-            space="{space}",
-            hemi="{hemi}",
-            label="{autotop}",
-            suffix="surfaces.spec",
-            **config["subj_wildcards"],
-        ),
-        density=config["output_density"],
-        space=ref_spaces,
-        hemi=config["hemi"],
-        autotop=config["autotop_labels"],
-        allow_missing=True,
-    )
+    if len(config["hemi"]) == 2:
+        specs = expand(
+            bids(
+                root=root,
+                datatype="surf",
+                den="{density}",
+                space="{space}",
+                label="{autotop}",
+                suffix="surfaces.spec",
+                **config["subj_wildcards"],
+            ),
+            density=config["output_density"],
+            space=ref_spaces,
+            autotop=config["autotop_labels"],
+            allow_missing=True,
+        )
+    else:
+        specs = expand(
+            bids(
+                root=root,
+                datatype="surf",
+                den="{density}",
+                space="{space}",
+                hemi="{hemi}",
+                label="{autotop}",
+                suffix="surfaces.spec",
+                **config["subj_wildcards"],
+            ),
+            density=config["output_density"],
+            space=ref_spaces,
+            hemi=config["hemi"],
+            autotop=config["autotop_labels"],
+            allow_missing=True,
+        )
     return specs
 
 
--- a/hippunfold/workflow/rules/common.smk
+++ b/hippunfold/workflow/rules/common.smk
@@ -24,41 +24,69 @@ def get_modality_suffix(modality):
 
 
 def get_final_spec():
-    if len(config["hemi"]) == 2:
-        specs = expand(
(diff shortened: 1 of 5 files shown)


jordandekraker:
@akhanf please merge https://github.com/khanlab/hippunfold_deps/pull/12 before reviewing this PR

akhanf:
Great! I'll try to have a look soon, and also build the design container (lock file needs updating first I think).

hippunfold/workflow/rules/common.smk
```diff
@@ -24,41 +24,69 @@ def get_modality_suffix(modality):
...
-        specs = expand(
+    specs = expand(
+        bids(
+            root=root,
+            datatype="surf",
+            den="{density}",
+            space="{space}",
+            hemi="{hemi}",
+            label="{autotop}",
+            suffix="surfaces.spec",
```
akhanf, line 35: previously the surfaces spec would have both hemispheres in it - this seems to change it so we have a spec file for each hemisphere? was that intended?

hippunfold/workflow/scripts/create_warps.py
```diff
@@ -104,74 +97,63 @@ def summary(name, array):
...
 # we have points defined by coord_flat_{ap,pd,io}, and corresponding value as native_coords_phys[:,i]
 # and we want to interpolate on a grid in the unfolded space
 
-# add some noise to avoid perfectly overlapping datapoints!
-points = (
-    coord_flat_ap + (np.random.rand(coord_flat_ap.shape[0]) - 0.5) * 1e-6,
-    coord_flat_pd + (np.random.rand(coord_flat_ap.shape[0]) - 0.5) * 1e-6,
-    coord_flat_io + (np.random.rand(coord_flat_ap.shape[0]) - 0.5) * 1e-6,
-)
+# unnormalize and add a bit of noise so points don't ever perfectly overlap
```
akhanf, line 103: doesn't seem like noise is being added anymore - perhaps can remove the comment if so

hippunfold/workflow/scripts/create_warps.py
```diff
@@ -104,74 +97,63 @@ def summary(name, array):
...
-)
-summary("interp_ap", interp_ap)
-interp_pd = griddata(
-    points, values=native_coords_phys[:, 1], xi=unfold_xi, method=interp_method
-)
-summary("interp_pd", interp_pd)
-interp_io = griddata(
-    points, values=native_coords_phys[:, 2], xi=unfold_xi, method=interp_method
+# perform the interpolation
+
```
akhanf, line 128: actually looks like noise is added here - so just need to move the comment

jordandekraker: Actually, `unfolded_gx` `unfolded_gy` `unfolded_gz` were no longer used, so I removed them. This also means that epsilon is no longer used, and i believe its no longer needed

jordandekraker:
OK i cleaned up the reviews. One more question: does https://github.com/khanlab/hippunfold_deps build the hippunfold_deps container automatically or do i need to trigger it?

akhanf:
I've made a hippunfold_deps release (v0.5.1) which triggers the docker hub build. Will need to update the container version here to point to that too.

PR #307, opened 2024-06-27
patch-templateseg-dHCP

jordandekraker:
Missed this in https://github.com/khanlab/hippunfold/pull/304 (caught in wet-run)


jordandekraker:
Relatedly, I now notice that the Dice scores are missing from all `--use-template-seg` workflows. I'll see about patching this bug as well.

jordandekraker:
Come to think of it, Dice from `--use-tempalte-seg` would be meaningless since both measures are computed the same way. The only difference is that one uses `reg_f3d` while the other uses `greedy` for registration

PR #310, opened 2024-07-10
Template inject Lflip vs R

jordandekraker:
I noticed that when using `--inject_template bigbrain`, only the hemi-R template was being used (with the hemi-L segmentation image) when actually both L and R are available from this template. I fixed this with the same logic as the new `templateseg.smk` logic.

Waiting for me new wetrun cases to finish


jordandekraker:
wetrun cases working

PR #318, opened 2024-10-29
Update snakebids to its latest version (v0.13.1)

Dhananjhay:
This PR updates hippunfold to use the latest version (v0.13.1) of snakebids.


akhanf:
I've changed the base of this to merge into (a newly created) dev-v2.0.0 branch, which we can use for this and any other features/breaking-changes we want for v2.0.0

so hoping to merge this PR soon once I give it a whirl!

hippunfold/run.py
```diff
@@ -1,24 +1,27 @@
 #!/usr/bin/env python3
-import os
+from pathlib import Path
 
-from snakebids.app import SnakeBidsApp
-from snakebids.cli import add_dynamic_args
+from snakebids import bidsapp, plugins
 
+if '__file__' not in globals():
```
akhanf, line 6: not sure why this is needed? it worked fine without for me?

Dhananjhay: I had to add this to pass the readthedocs check; it has more to do with the documentation than the application itself.

akhanf: ah ok then we can leave it as you have it..

akhanf:
ok I've done a bit of refactoring on the expands() and reading bids inputs. I did also have to move some rules out of the preproc_t1.smk and into warps.smk (they were rules for dealing with transforms that were needed for either T1w or T2w), since I had to make the preproc_t1 rule file only included if we are grabbing T1w data (as now snakebids would give a key error otherwise when trying to do inputs["T1w"].

akhanf:
@jordandekraker, @Dhananjhay after you guys have had a look I think this one is ready to merge in, then more PR's can be made into the dev-v2.0.0 branch..

jordandekraker:
looks good to me! I say we merge and then i can start some local tests

jordandekraker (approved):
I have not tested locally, but am eager to start development on the 2.0.0 branch

Dhananjhay:
Yes, looks good to me! I can try running some tests locally too.

akhanf:
Great I'll merge this in, I did dry runs and some wet runs but if there is anything to iron out we have time to catch it in a later PR since this isn't being merged into master yet.

PR #325, opened 2024-12-12
simplified use of manual segmetation

jordandekraker:
I've had many users get tripped up on this (myself included!), often when using ex-vivo data. Here, I  provide a single `--modality` option for manual segmentations. I add the requirement that the data be roughly aligned with one of the input templates (e.g. CITI168). I believe this is a simple requirement given the complexity anyone has to go through in doing a manual segmentation anyway. Also ex-vivo cases tend to be in arbitrary spaces anyway, making it very hard to work with until they are reoriented.

I also restore the lut, which I will ad to the Documentation guiding manual segmentations (TODO!)

The template is still used to go to space-corobl (with cropping), so users don't need to crop around the hippocampus. So, effectively, I'm renameing `cropseg` to `manualseg`, adding the template-to-corobl transform, and remove the `segT1w` and `segT2w` workflows.


jordandekraker:
check should pass with https://github.com/khanlab/hippunfold/pull/326/

jordandekraker:
double check common.smk

PR #326, opened 2024-12-12
Ali fixes to dev branch

jordandekraker:


jordandekraker:
@akhanf I think I got your updates to master resolved here correctly, but can you please confirm?

akhanf:
The config file changes aren't needed (the bids args are now added automatically), but otherwise looks fine I think.

akhanf:
@jordandekraker looks like you (accidentally?) merged dev-2.0.0 into master.. 
we should undo that as we wanted to reserve this until 2.0 release candidate is ready, so we don't break things on master along the way.. I can do a force-push to undo the changes, but then would want to make sure you don't use any branches since that merge (~5pm sunday)

jordandekraker:
I think this way is correct. The scope of this PR is to merge commits from master into 2.0.0, but there were merge conflicts so i needed that last commit to resolve them. Since the 2.0.0 changes are newer than master, github automatically names this merge as 2.0.0 into master but actually its still within the scope of 2.0.0 (no changes to master).

The funny thing is, i can't take out the config changes from that merge commit since there's not an actual branch for master that's within this 2.0.0 😆 i'll roll back the merge and try again

jordandekraker:
I was wrong, my merge got pushed to master 😞 
We'll clean it up as discussed via zoom

PR #327, opened 2024-12-13
Nativesurf

akhanf:
This adds generation of surfaces in native space (isosurf on IO coords), and then uses resampling with wb_command tools, using the unfold surfaces as one would use "sphere" surfaces in HCP.. 

Few things we will want to address in future PRs:
- holes can exist in areas of thin cortex -- my initial fix to extrapolate (using the prep_hipp_coords_for_meshing rule) was leading to issues in other parts of the surface (points sticking out where they shouldn't), so I've disabled that for now. Could try a more constrained extrapolation? or hole-filling (might work best in unfold domain)...
- I haven't integrated this with unfolded_reg yet, as I wanted to refactor that a bit anyways.. so for now it disables it if --native-surf is used
- with this, we don't need to estimate inverse warps (ie to take surfaces from unfold to native), so those parts of the workflow can be removed
- inner and outer surfaces are created by warping the midthickness, using a new warp defined from volumetric mask registration -- ideally this should use the coords (e.g. propagate along streamlines)
- should consider a naming change, to introduce an "unfold" or "unfolded" suffix (alongside midhickness, inner, outer) for surf.gii files. Then we can use the `space` entity to denote an atlas, e.g. space=multihist7, space=bigbrain. Density would be used as is. 
- dentate surfaces are really holey.. using a higher-res resampling for dentate and higher-res template may be the way forward..


jordandekraker:
Pull https://github.com/akhanf/hippunfold/pull/1 for latest updates

I think that we should try to merge this branch soon and then work on refinements & updating the use of warps.smk and the Laplace–Beltrami for coords

jordandekraker:
don't rush on my behalf, i've already started another branch haha

PR #328, opened 2024-12-13
BB templateseg & outlier vertices

jordandekraker:
This is an older branch which I forgot to merge, but I think it is doable.

Basically the idea is to have a BigBrain template available in the templateseg workflow. This isn't always advisable, because of the well known issue of volumetric registration smearing gyri/sulci, but it can give a reasonable starting point for someone wanting to do their own manual segmentations or just to get an idea of how volumetric registration WOULD perform. 

I'm merging this into `master` rather than `dev-v2.0.0` since it relaly does pertain to the templateseg workflow. Will run a local test too.


akhanf:
ok hold off on merging -- I'm about to fix the accidental push ..

hippunfold/config/snakebids.yml
```diff
@@ -358,7 +365,13 @@ cifti_metric_types:
     - gyrification.dscalar
     - curvature.dscalar
 
-
+# params for detecting and removing outlier vertices
```
akhanf, line 368: would be nice to put these under a common key, e.g. outlier_opts.. 
when the config file gets rewritten to and output folder all the ordering gets lost and so having things grouped is a good practice.. (generally something I think we should add an issue for in v2.0)

jordandekraker:
Yes this still needs testing

I see that I also made some changes to try and correct bad vertices. I believe this was resolved in https://github.com/khanlab/hippunfold/releases/tag/v1.5.0 though, so its probably not necessary. I will remove those changes and then try a test.

hippunfold/workflow/rules/common.smk
```diff
@@ -24,41 +24,69 @@ def get_modality_suffix(modality):
...
+        density=config["output_density"],
+        space=ref_spaces,
+        hemi=config["hemi"],
+        autotop=config["autotop_labels"],
+        allow_missing=True,
+    )
+    return specs
+
+
+def get_final_surf():
```
akhanf, line 47: is this redundant if the surfaces are in the spec?

hippunfold/workflow/scripts/fillbadvertices.py
```diff
@@ -0,0 +1,59 @@
+import nibabel as nib
```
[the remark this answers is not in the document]
jordandekraker: i believe it was more than a rename, since it also tried to do outlier detection. I will test and then consider restoring the previous behaviour though, since this may not be needed

hippunfold/workflow/rules/common.smk
jordandekraker: I believe I added this so that we could specify mid-inner-outer surfaces in the dentate within the config file too. Without this they will never be generated since they are not in the spec. I will see with some testing though...

akhanf: I guess the better place to update it is in the expand() for the spec file then, ie if inner outer surfaces are requested then, they are put into the spec

jordandekraker: right, i remember now. This was a bit challenging since the config can be different for hipp and dentate surfaces.  This is likely solvable with wildcards, but I found a rule for surfaces to be a bit more intuitive since they are a major class out outputs, so its nice to have them listed explicitly. What do you think?

akhanf: would be nicer to have fewer target rules generally, but the main issue I worry about is having redundant code that gets out-of-sync, e.g. if we expand over surfaces in the cifti creation *and* the target rules -- the code so far would make it out of sync..  The `create_spec_file_dentate` rule is where you could just update what wildcards to expand over what you need (and then only have it appear once in the code).

akhanf:
just a heads up - still working on fixing the master branch, as it seems now the commits to merge the bugfix PRs are gone.. I made a backup just in case..

akhanf:
> just a heads up - still working on fixing the master branch, as it seems now the commits to merge the bugfix PRs are gone.. I made a backup just in case..

OK master should be good now (I was reverting back to the wrong commit!)

hippunfold/workflow/rules/common.smk
jordandekraker: was just looking to undo this when i found that it had actually already made its way into the master branch:

https://github.com/khanlab/hippunfold/commit/c32bac2a3369d2e8e77137e09b3db36f983ac19e#diff-44c3321b981bf8e9ad7e5f10740de98858b868cd8a75cbbcbbe08170249441ee

I'm not quite sure what you mean about these files going out of sync though

hippunfold/config/snakebids.yml
jordandekraker: [003dbf3](https://github.com/khanlab/hippunfold/pull/328/commits/003dbf36773548c6ebfd3df490cbe6f0e868c2e6)

jordandekraker:
This branch is technically working, BUT:

Major:
- Whole-brain registration almost always fails. This could be ameliorated by using greedy rigid registration (since we plan to remove nifty_reg anyway), but it could also potentially be ameliorated by including the phantoms from https://www.sciencedirect.com/science/article/pii/S1053811923002203. 

Minor:
- in generating warps, unfolded coords are now 256x128x16 instead of 1x1x1 when generating warps. This affects the behaviour of natural interpolation, but I believe its actually better.
- relatedly, epsilon is now expressed in unfolded voxels rather than as a fraction of coords
- outlier removal on surfaces is applied
- we have `def get_final_surf():` in `common.smk`, but this is actually also in the master branch

hippunfold/workflow/rules/common.smk
akhanf: ok then it's something we can clean up in the v2.0 refactor -- I wasn't talking about files going out of sync, but about redundancy in the code (eg expanding over wildcards in two different places, and having to maintain those two expands to be the same..) - probably occuring more than just here, but is something we should try to avoid for easier maintenance (ie in the future when we change things, only have to change in one place, not two..)..

jordandekraker:
was thinking about this during lunch: given the magnitude of the changes, maybe we should merge this to dev-v2.0.0.

Also to consider: with the new native surfaces we may end up changing the behaviour of `warps.smk` completely, so having this in the dev branch might be best. @akhanf does that sound good? 

I'll see about adding those phantoms and if the registration runs well now.

PR #332, opened 2024-12-16
Conda dependencies

Dhananjhay:
This PR removes the `equivolume` logic from the laminar coordinate method, effectively eliminating hippunfold's dependency on the `nighres` package. This change serves as a stepping stone toward synchronizing and integrating hippunfold with the conda environment.

### DIFF
--- a/hippunfold/config/snakebids.yml
+++ b/hippunfold/config/snakebids.yml
@@ -255,12 +255,8 @@ parse_args:
     help: 'Hemisphere(s) to process (default: %(default)s)'
 
   --laminar_coords_method:
-    choices:
-      - 'laplace'
-      - 'equivolume'
-    default:
-      - 'equivolume'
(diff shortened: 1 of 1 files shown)


Dhananjhay:
@akhanf I've only removed `equivolume` as a choice from the `config/snakebids.yml` but I was wondering If I should clean the code-base more extensively. For example, I believe we no longer need `scripts/equivolume_coords.py` and `scripts/pre_equivolume_coords.py`and the relevant logic in `wraps.smk`, `common.smk` and `autotop.smk`. Moreover, I can edit `algorithms.md` and `pipeline.md` to reflect that equivolume is no longer offered as a choice in `laminar_coords_method`.

akhanf:
We haven't decided yet what route we will take so no need to clean up yet. But this should at least allow you to progress on the pure conda hippunfold

Dhananjhay:
I don't think this PR is relevant anymore. For more details, please follow PRs #351 and #349.

Dhananjhay:
I don't think this PR is relevant anymore. For more details, please follow PR #349 and #351.

PR #333, opened 2024-12-17
BB template & vertex outliers

jordandekraker:


jordandekraker:
Merging `dev` into `BBtemplate` I began to remove the redundant `get_final_surfaces()`, but this will likely take some tweaking still

hippunfold/workflow/scripts/create_warps.py
```diff
@@ -100,23 +100,30 @@ def summary(name, array):
...
+    + (np.random.rand(coord_flat_ap.shape[0]) - 0.5) * 1e-6,
+    coord_flat_pd * unfold_dims[1]
+    + (np.random.rand(coord_flat_ap.shape[0]) - 0.5) * 1e-6,
+    coord_flat_io * unfold_dims[2]
+    + (np.random.rand(coord_flat_ap.shape[0]) - 0.5) * 1e-6,
+)
+
+# get unfolded grid (from 0 to 1, not world coords), using meshgrid:
+#  note: indexing='ij' to swap the ordering of x and y
+epsilon = snakemake.params.epsilon
```
akhanf, line 115: I know this PR is merged already, but easier to comment here than anew -- these changes break the rule because epsilon is not a parameter in the rule. Also, unfold_gx/unfold_gy/unfold_gz get create here, but don't actually get used .. so it seems we can just revert the create_warps changes to fix the problem, unless you were trying to do something else here?

jordandekraker: oh yeah, good point. Looks like we were never actually using the epsilon that we'd previously defined? I'll make a small PR to patch this & 1 other small bug i found

Also yeah `points` and `unfold_gx` somehow became redundant. We can cut the latter.

PR #335, opened 2024-12-17
Simple lflip

jordandekraker:
Here we remove all instances of flipping EXCEPT for templateseg and and shape_inject (where we now apply them to the templates instead of the images)


jordandekraker:
wetrun testing right now

jordandekraker:
My wetrun test seems to have passed (I also modified the test to handle the new manualseg)

Note that the templateseg neonate pipeline doesn't look great (the registration seems to sometimes place the hippocampus in the wrong place), and it may even have performed better using nnunet.

I think this PR is ready for merge!

hippunfold/config/snakebids.yml
```diff
@@ -356,37 +356,28 @@ xfm_identity_itk: resources/etc/identity_xfm_itk.txt
 
 #templates enabled for template-based segmentation are here
 # TODO: should also perhaps include modalities avaialble, and any custom crop_native_res settings
-template_based_segmentation:
+template_available_hemis:
```
akhanf, line 359: Why not put this inside the template_files dict to keep things together? I can do that in my edits..

akhanf: just made that edit -- generally trying to keep things hierarchical in the config (will make anotehr issue to do this en-masse, as the config right now is really messy with so many different top-level keys (which don't remain sorted in the nice way that we have them when deployed to an output folder)..

akhanf:
this is good to merge from my end -- I think you could do a Squash and merge instead,  to keep the history clean (since all the commits can conceptually fit in a single commit)

PR #342, opened 2025-01-03
Laplace-Beltrami on surfaces

jordandekraker:
Using the Laplace-Beltrami operator on a surface is much faster than a finite-differences solution on a volume, especially at very high resolutions. 

Eventually I would like to apply this to levelset surfaces generated directly by nnunet to avoid having to use the Laplace solver entirely. This has the secondary advantage of solving "holes" and "bridges" in the SRLM and alveus segmentations, allowing us to eliminate `shape_inject.smk`. Using Laplace-Beltrami on top of this would be a large speed boost, but even keeping just the IO Laplace solver is still an improvement since AP is the slowest to solve (furthest distances). 

Notes: 
- native surfaces should have no holes, and must have only one connected-component
- the current surface edge detector is sensitive to spikey edges

I hope to clean up both of these issues by playing around with levelsets

This logic is shown in one notebook (for illustration only). This should eventually replace `autotop.smk` (possibly keeping the IO equivolumetric solver) and `warps.smk`


akhanf:
Cool looks promising! I also agree that IO coords volumetrically then LB on surface could be a good way forward. 
 
Probably lots to catch up on in Tues call - I have a PR that has improvements to native surf along with refactored unfolded registration - hipp surfaces are looking great now with no holes or adhoc fixes, and should work well with this. PR is just about ready hope to put a draft up this weekend.

jordandekraker:
taken over by https://github.com/khanlab/hippunfold/pull/353

PR #346, opened 2025-01-10
manualseg resampling in-place without transformation

akhanf:
This PR adds an alternative behaviour for modality=manualseg, where the segmentations are not transformed at all, but just resampled in place to alter the resolution if needed. I used this e.g. to run hippunfold on the bbhist manual segmentations. 

Could think about the implementation of the CLI for this behaviour, maybe specifying the resolution as mm instead of % would make more sense, maybe could co-opt the crop_native_res resolution arg, and make this one a boolean? 

@jordandekraker wondering if we should make this option the default behaviour for manualseg - only downside I see is that the outputs are named corobl, when they are not really corobl..


jordandekraker:
One initial concern I had with this is that the Laplace solver benefits hugely from tight cropping, but with Laplace-Beltrami instead that shouldn't be an issue. 

The other thing I would worry about is whether tempalte-shape injection will fail (people often show me manual segmentations in images that are completely disoriented, e.g. upside down because its was easier to scan a specimen that way). Since manual segmentations are pretty variable in quality, this template-shape injection can really help lower the bar for required quality. This could be a non-issue with moment-of-intertia-based initialization of the registration though.

Probably the best way to know is with a few test-cases. I'll add a few samples to the wetrun manual seg with random rotations (with and without correpsonding headers), and we'll see how it does

akhanf:
Re: cropping that is a good point, we should actually add a cropping step here too (could be in the same c3d call)

For shape injection yes would be good to test that out more, I didn't have it enabled when I used it on bbhist.

akhanf:
Yeah, will need to use the --path-manualseg option along with --hemi.. e.g.:

```
hippunfold test_data/bids_manualseg test_out participant -np --modality manualseg --path-manualseg test_data/bids_manualseg/sub-{subject}/anat/sub-{subject}_hemi-{hemi}_dseg.nii.gz --hemi L R
```

akhanf:
OK i'm gonna make a couple minor updates to this (remove the previous corobl resampling, make in-place resampling optional, and add a step to crop tight (with limited padding) around the segmentation. 

We could also add in an optional reorientation step (to deal with really wonky orientations, e.g. where the user specifies an input orientation string) - @jordandekraker let me know how your tests work out and if that is needed for e.g. template injection to work well..

akhanf:
OK made those changes - should be ready to go

jordandekraker:
ran a few tests with the bigbrain manual segementation with random rotations (and scrambled header info). The template injections works great when the template shape is also bigbrain, but when the the template shape is the default upenn, it worked in one rotated case and failed in another. Overall I think this is fine and we can merge this PR, but there's always a possibility that with weirdly rotated input images, the momentum and/or affine registration to the template could be way off (e.g. upside down). Luckily, the usage text still notes that its best if the input is in a standard orientation, which I think we can leave and merge this!

PR #349, opened 2025-01-13
WIP - Laynii native

akhanf:
Tried this out some time last week, but posting here as a WIP since @Dhananjhay and @mackenziesnyder mentioned they will have a look at it too..


akhanf:
OK this is good to merge now -- note: unfolding for dentate won't work until the templates are updated with the proper dentate tissue src/sinks.. Also, I have rules for the renzo versions of equivol and equidist, but I haven't included them in the workflow (nor have I tested them), but they are there as a starting point if we wanted to check them out..

PR #360, opened 2025-01-30
Upgrade snakebids to latest version

Dhananjhay:
Everytime hippunfold is run, the terminal spits out numerous warnings regarding the use of relative paths, i.e.,
```
Relative file path './sub-01/surf/sub-01_space-T1w_den-0p5mm_surfaces.spec' starts with './'. This is redundant and strongly discouraged. It can also lead to inconsistent results of the file-matching approach used by Snakemake. You can simply omit the './' for relative file paths.
Relative file path './sub-01/surf/sub-01_space-T1w_den-2mm_surfaces.spec' starts with './'. This is redundant and strongly discouraged. It can also lead to inconsistent results of the file-matching approach used by Snakemake. You can simply omit the './' for relative file paths.
Relative file path './sub-01/surf/sub-01_space-T1w_den-unfoldiso_surfaces.spec' starts with './'. This is redundant and strongly discouraged. It can also lead to inconsistent results of the file-matching approach used by Snakemake. You can simply omit the './' for relative file paths.
Relative file path './sub-01/surf/sub-01_space-corobl_surfaces.spec' starts with './'. This is redundant and strongly discouraged. It can also lead to inconsistent results of the file-matching approach used by Snakemake. You can simply omit the './' for relative file paths.
```
It was getting harder to debug and read the execution of the pipeline from the terminal therefore I've added this small patch to fix things.


akhanf:
I've actually fixed this in snakebids (https://github.com/khanlab/snakebids/pull/462) to remove the leading dot.

Not a good idea to use absolute paths for snakemake paths (e.g. makes defining targets at the command-line tedious) --  can you check if upgrading snakebids to 0.14.0 fixes it without your change to absolute paths?

Dhananjhay:
Yes, no warnings when upgraded `snakebids` to version `0.14.0`. Awesome, thanks @akhanf!

akhanf:
feel free to upgrade in this PR

akhanf (approved):

PR #364, opened 2025-02-05
add decimation and clean_method params for gen_isosurface

akhanf:
- for clean_method, cleanAK and cleanJD can be specified, but in my hippocampal test case, they are indistinguishable..
- decimation is also enabled by default to fix memory issues with surface resampling (even just a little decimation seems to solve it)
- note: decimation only works with cleanAK (cleanJD seems to choke on decimated surfaces)..

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,6 @@ dist/
 downloads/
 eggs/
 .eggs/
-lib/
 lib64/
 parts/
 sdist/
--- a/hippunfold/config/snakebids.yml
+++ b/hippunfold/config/snakebids.yml
@@ -688,6 +688,7 @@ laplace_labels:
         - 8
       src:
         - 1
+        - 3
       sink:
         - 2
         - 4
--- a/hippunfold/config/snakebids.yml
+++ b/hippunfold/config/snakebids.yml
@@ -660,7 +660,6 @@ laplace_labels:
     IO:
       gm:
         - 1
-        - 8
       src:
         - 2
         - 4
--- a/hippunfold/workflow/Snakefile
+++ b/hippunfold/workflow/Snakefile
@@ -120,26 +120,12 @@ if config["modality"] == "hippb500":
 
 
 include: "rules/coords.smk"
-
-
 include: "rules/warps.smk"
-
-
 include: "rules/gifti.smk"
-
-
 include: "rules/subfields.smk"
-
-
 include: "rules/resample_final_to_crop_native.smk"
-
-
 include: "rules/qc.smk"
-
-
 include: "rules/myelin_map.smk"
-
-
 include: "rules/native_surf.smk"
 
 
--- a/hippunfold/workflow/Snakefile
+++ b/hippunfold/workflow/Snakefile
@@ -120,12 +120,26 @@ if config["modality"] == "hippb500":
 
 
 include: "rules/coords.smk"
+
+
 include: "rules/warps.smk"
+
+
 include: "rules/gifti.smk"
+
+
 include: "rules/subfields.smk"
+
+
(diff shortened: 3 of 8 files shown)


akhanf:
actually, it looks like even with decimation reduction of 0 (ie no actual decimation), the workbench crashing issue is resolved, so likely the problem has to do with something strange happening to the surfaces in cleanJD..

akhanf:
Sorry I actually have some additional uncommitted changes, using pyvista for cleaning, hole filling, connected comps (all one liners), will merge when I have a chance.. 

Will see if I can do the equivalent morphological operation with built-ins too, also experimenting with smoothing to clean that edge..

akhanf:
Ah also re: decimation, the reduction is not in terms of physical spacing but in proportion of total number of vertices to remove, so can't specify that with mm.. anyways will clean up a bit with my merge

akhanf:
OK for bbhist things are looking nice in the hippocampus for me (so far with cleanJD off), will have a look at dentate next (maybe later tonight or tomorrow..)

Btw, any reason we have been including the DG (label 8) in the domain when calculating IO coords in hippocampus? To me things look more anatomically correct when using just label 1 (hipp GM, not including dentate).. I know it's been 1+8 for awhile now but I can't recall the motivation, maybe was to overcome other issues with nighres? Anyways, have a look yourself if changed to just label 1..

jordandekraker:
cool, i'll give it a try! just keep an eye out for holes if they pop up in dentate gyrus (or in hipp). hopefully pyvistas fills these

re: including label 8, at one point we were estimating dentate PD coords as being hipp IO coords (a heuristic that works OK since they are roughly perpendicular). It would be reasonable to exclude it from the hipp pipeline!

akhanf:
Haven't got to the dentate yet, but here's an example of cleanJD (left) and cleanAK (right) for the current code, on bbhist left hipp, 

image


also zooming in on the one defect I could find.. 
image

akhanf:
ok surface generation with dentate is looking good now too so I think we can close this PR. Note there are two holes (with both cleanAK and cleanJD) that seem to get mostly patched up with hole-filling.. if they persist we can perhaps try to deal with them in another PR.. 

NOTE: the laplace-beltrami A-P for dentate is having a hard time since the labels don't overlap with the vertices (off by a little bit), had this issue when dealing with injected labels earlier too -- I have an idea to simply label the vertex closest to the src and closest to the sink (that should be robust, what do you think?), will implement that in another PR though..

akhanf:
Here are some snaps. Looks like we can up the pre-smoothing later for curvature estimation, as the decimation might be adding a bit more noise. Or add a tiny bit of smoothing to the actual surfaces (none enabled right now), the smooth_taubin() function is pretty customizable and is designed to avoid vertex drift, can optimize in a future PR (will post an issue).. 

image
image
image
image

akhanf:
Note: only showing thickness and curvature, until we update the gyrification measure.. will make an issue for that too..

jordandekraker:
Sounds good! I'll be running the multihist7 today (as long as my segmentations prove good enough), hopefully there are no issues but if so i can post them to the new smoothing issue.

Not quite sure what you mean about laplace-beltrami labels overlapping vertices. This is the case for hipp surfaces as well, and the idea was to simply get the nearest neighbour between APsrc|APsink|PDsrc|PDsink for every edge vertex. The main problem I've noticed is that if there is a hold in the surface, those vertices are labeled as an edge an assigned a src|sink. A second issue I saw in the bbhist hemi-R is that part of the DG is so folded that the sink appears to be halfway along the width of the DG (but the DG is actually folded on itself). This can lead to funny looking surfaces, but I might still clean it up with my manual labeling

PR #365, opened 2025-02-06
New participant_create_template analysis_level

akhanf:
Adds a target rule for generating the files for using a subject as a surface template, i.e. just generating the native and unfold surfs, and metrics and subfields (i.e. everything required for an `atlas:` entry in the config file)..  

This is set-up as a new analysis-level, participant_create_template, which we would run for our multihist7 subjects. Could also use this in the future to make templates from MRI data, so one could get e.g. myelin maps or rsfmri connectivity as metrics (which could then be used in unfoldreg when a subject has that data).. 

TODO: add sampling of manual volumetric segmentation into surface metric

Note: this includes commits from #364


akhanf:
@jordandekraker you can use this branch for your multihist7 tests, just run `participant_create_template` instead of `participant`, and it will automatically skip template injection and only generate the files needed to create a template (e.g. will skip unfoldreg).. 

I still need to add the rules to load and resample manual volumetric subfield labels with the native surfaces, but otherwise is functional.

jordandekraker:
I'm having an issue which I believe comes from https://github.com/khanlab/hippunfold/pull/364

```
Traceback (most recent call last):
  File "/export03/data/opt/hippunfold-dev/multihist7_hippunfold/.snakemake/scripts/tmpc8fix6to.gen_isosurface.py", line 11, in <module>
    from lib.utils import setup_logger
ModuleNotFoundError: No module named 'lib'
```
I suspect I need to add this new directory to my path (maybe in the top level Snakefile?)

akhanf:
You should have `hippunfold/workflow/lib/utils.py` in this branch too, does it not exist for you? even after pull? 
It seems to be there for me.. (and snakemake should find it this way)..

jordandekraker:
yes its there, just not being seen by python scripts

jordandekraker:
solved by using conda install (pip and poetry are currently broken):
```
conda create -c conda-forge -c bioconda -n snakemake snakemake
conda activate snakemake
conda install -c conda-forge -c bioconda snakebids
./hippunfold/run.py
```

akhanf:
This is good to merge from my end, later I'd like to add in some logic to prepare the manifest and data files for registering a new template (ie commit to a repository, which could be your own repo, or the centralized one), but details of that in a future PR..

PR #371, opened 2025-02-09
Refactor laplace-beltrami 

akhanf:
This PR deals with the steps leading up to the laplace-beltrami solver, not the solver itself (that is kept as is)

- splits into different rules for extracting the boundary, and getting distances to src/sink masks
- those new rules now produce surface maps (for easier introspection), which are passed along to the laplace beltrami rule
- the dentate AP src in bbhist hemi-L was problematic with existing approach, since the PD src/sink were always closer than the AP src, so no vertices were being labelled with AP src.
- in this version, we do things a little different:
  1. AP and PD are done independently now (this also helps with computational efficiency, since they can run in parallel)
  2. Instead of nearest neighbor, we use a distance threshold, which is defined based on a signed distance transform of the mask, mapped to the surface.
  3. Actually there is a minimum distance, and maximum distance, along with the minimum number of vertices required in the src/sink. The threshold is swept until the desired number of vertices is reached, to deal with cases such as the above (throws an error if not reached).

A couple maybe todo's:

- TODO: the get_boundary_vertices script could be further improved to perform connected components and picking the largest one, to avoid being affected by holes or other defects..

- TODO: the laplace beltrami solver doesn't take as long to run now that I am decimating alot, but if we want to further optimize things, I think the step of altering the laplacian based on the boundary conditions, after it has been made a sparse matrix, is where the inefficiency lies (based on profiling). Could try to avoid setting those weights in the first place instead of setting it after the fact..

### DIFF
--- a/hippunfold/workflow/rules/coords.smk
+++ b/hippunfold/workflow/rules/coords.smk
@@ -21,21 +21,13 @@ def get_gm_labels(wildcards):
     return lbl_list
 
 
-def get_sink_labels(wildcards):
+def get_src_sink_labels(wildcards):
     lbl_list = " ".join(
         [
             str(lbl)
-            for lbl in config["laplace_labels"][wildcards.label][wildcards.dir]["sink"]
-        ]
-    )
-    return lbl_list
-
-
-def get_src_labels(wildcards):
-    lbl_list = " ".join(
-        [
-            str(lbl)
-            for lbl in config["laplace_labels"][wildcards.label][wildcards.dir]["src"]
+            for lbl in config["laplace_labels"][wildcards.label][wildcards.dir][
+                wildcards.srcsink
+            ]
         ]
     )
     return lbl_list
@@ -100,11 +92,11 @@ def get_inputs_laplace(wildcards):
     return files
 
 
-rule get_sink_mask:
+rule get_src_sink_mask:
     input:
         labelmap=get_labels_for_laplace,
     params:
-        labels=get_sink_labels,
+        labels=get_src_sink_labels,
     output:
         mask=bids(
             root=work,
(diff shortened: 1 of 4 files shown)


akhanf:
> This makes sense to me, I even had a note about possibly separating this into two rules. The issue of dentate not reaching all the way to anterior and posterior is one that i solved by editing manual segmentations in mutihist7, but this will certainly make the code more robust.

I was using the bbhist you recently posted (maybe was updated since) but yes I think having ways to make it more robust is a good idea (esp when we apply to lower-res, ML/injected images). 

> 
> Just two questions:
> 
>     * is there any mechanism to ensure that a vertex isn't assigned to a terminus in both AP and PD solutions? I think that ideally this would occur so that the two coords are closer to orthogonal, but maybe not strictly necessary since it would need to be done in serial rather than parallel

Yes I was thinking about this too, but didn't tackle it here as the laplace beltrami rule now only as one set of src/sink.  We could do this by adding a rule just upstream that takes in both AP sdt, PD sdt (and boundary mask), and outputs boundary AP/PD masks. Then laplace_beltrami could take just one pair of masks (instead of all the sdt and boundary inputs). In this script we could have also some logic to ensure the src/sink masks are also one connected component (which is not always guaranteed, especially when dist threshold very low or very high).. Could fix the labels based on some heuristic or raise an exception.. 

> 
>     * why not make the minimum distance threshold 0? This would help in higher resolution manual seg cases, and could still occur if we had not removed label (DG) from the label-hipp foreground
> 
Yes we could, or maybe we make it proportional to voxel spacing? Once we have a check for connected components (above) then I think its fine to go low..
> 
> I agree on the TODO of adding conncomp to get_boundary_vertices, since this will make the code more robust to holes. Also speeding up the L-B solver sound good too, I noticed a warning about changing sparse matrix types which could be an easy fix.
> 
> Let me know if you'd like me to take on these TODOs, but I won't be able to run until my ubuntu finishes updates sometime this afternoon

I'm inclined to just merge this in (since it's functional, and other PRs are in line), and make issues for these other points..

jordandekraker (approved):

PR #374, opened 2025-02-11
Updates for shape injection and surf processing of injected surfaces

akhanf:
This PR makes changes to injection and surface processing for dealing with native surf with nnunet + injection.


akhanf:
updated upenn dseg is uploaded to osf now, so I will merge this now to move things along.. 

Will see if I can try to get the updated src/sink labels and new gyrification done before I am travelling tomorrow..

PR #379, opened 2025-02-21
Atlasgen

jordandekraker:
Here we will try to re-do template (atlas?) generation in the multihist7 dataset. This entails pooling surface metric files from all subjects. It willl probably make sense to put the atlas in a new output folder at the top level. For now i'm using "mytemplate" as the default name, but we should also think about what optional parameters we want the user to use for atlas generation and whether they should specify with CLI arguments or instead edit the config directly. 

I also tried ot organize the rules a bit better by making `unfold_reg.smk` (no code updates, just moved some rules) and `atlas_gen.smk` which we will mostly be working on here (as well as some updates in the config and common.smk)


akhanf:
Just about to have a look now - at a glance, I think your problem is having expand in both input and output (you don't need the output expand).. but I also might refactor that rule a bit

akhanf:
Right now, this is using the ants bash script to perform template creation -- I am thinking this is probably fine since the ANTS 2D runs don't take much time, but we could move to a more granular workflow by taking rules from here: 
https://github.com/akhanf/ants_build_template_smk

akhanf:
I've made a couple minor fixes to get the dry-run working

akhanf:
this adds another fix to the target rules so you shouldn't have to run hemispheres separately.. 
I'll have a look at some of the other config/dag related things too, will try to commit often so we don't get out of sync..

akhanf:
ok I've committed some more changes to get the workflow working, up to and including grabbing the warps from the ants template generation.

akhanf:
hopefully the back and forth commits aren't disruptive -  I've just mainly been doing minor clean-ups and fixes along the way, so as long as you're pulling often should be fine. I'll have a look at things after your next commit again.. Getting closer! 

If I have time tomorrow, will try to work out a solution for template files (so far I've been removing references for the new template files in the config, since that would be circular), so can just continue putting things in the template folder for now until we sort that out.. also we can clean up naming later..

jordandekraker:
yeah I was working on the circular issue here. Basically I wanted to `--atlas` to be the files generated in the atlasgen. This is challenging though because it conflicts with downloads.smk. Since we had intended to enumerate the items in downloads.smk anyway, i thought I could solve it here (i.e. by writing all `### final atlas files ###` to the download_dir. We could conisider reverting that last commit though and doing it in a new branch instead, but i think i was getting close here...

I think we should also separate the `### final atlas files ###` files out to a new directory too, so its easier to upload to a cloud.

hippunfold/workflow/rules/atlas_gen.smk
```diff
@@ -0,0 +1,332 @@
...
+    script:
+        "../scripts/set_metric_nii_header.py"
+
+
+def get_metric_template(wildcards, input):
+
+    num_modalities = len(reg_metrics)
+
+    warp_prefix = "template/warp_label-{label}_"
+    metrics = sorted(glob(f"{warp_prefix}_template?.nii.gz"))
```
akhanf, line 229: the problem here is would be globbing before the files exist.. we can instead just point to the output files created in the `copy_avgtemplate_metric` rule. I'll update this..

hippunfold/workflow/rules/atlas_gen.smk
```diff
@@ -0,0 +1,332 @@
...
+            "..",
+            "resources",
+            "unfold_template_{label}",
+            "tpl-avg_space-unfold_den-{density}_midthickness.surf.gii",
+        ),
+    shell:
+        "cp {input} {output}"  # TODO get actual densities
+
+
+rule avg_metrics:
```
akhanf, line 277: Not sure what you're trying to average over here, as the template#.nii.gz files are already averages, each being a different channel  -- anyways I fixed this in my previous commits will revert back

hippunfold/workflow/rules/download.smk
```diff
@@ -3,18 +3,43 @@
...
-        "minimal"
-    shell:
-        "wget 'https://{params.url}' -O temp.zip && "
-        " unzip -d {output.unzip_dir} temp.zip"
+# rule download_extract_atlas:
+#     params:
+#         url=config["resource_urls"]["atlas"][config["atlas"]],
+#     output:
+#         unzip_dir=Path(download_dir) / "atlas" / config["atlas"],
+#         label_gii=Path(download_dir)
```
akhanf, line 11: removing the named outputs (label_gii etc..) might cause issues with the dag.. I'll see what can be done..

hippunfold/workflow/scripts/surf_gen.py
akhanf, line 1: should probably rename this (semantically too similar to gen_isosurface.py) -- what about smething liek: triangulate_avg_template?

jordandekraker:
Cool, i'll look into the downsampling. A lower res meshgrid would work, but i'm wondering if i should try to keep more vertices in higher-curvature areas (or according to the average native spacing). I don't think that there's any advantage to using a meshgrid since we can't reshape into APxPD anymore anyway, but it is also nice and simple

akhanf:
First look at a subject surface (1mm T1w) with standard mesh from the new avgatlas approach!

image

```
(snakemake) alik@AFI-CBS-H-19:/local/scratch/hippunfold_khanlab/ds002168_newatlas/sub-1425/surf$ wb_command -surface-information sub-1425_hemi-R_space-T1w_den-multihist7_label-hipp_midthickness.surf.gii
Name: sub-1425_hemi-R_space-T1w_den-multihist7_label-hipp_midthickness.surf.gii
Type: Anatomical
Number of Vertices: 28897
Number of Triangles: 56577
Bounds: (8.73535, 27.4752, 1.21742, 38.3933, -14.7856, 10.8633)
Spacing:
    Mean: 0.238097
    Std Dev: 0.151468
    Minimum: 0.000000
    Maximum: 1.824883
```

close to ~0.24mm vertex spacing, but becomes pretty oversampled in the uncus.. 


Note, the atlas unfold mesh had:
```
(snakemake) alik@AFI-CBS-H-19:/local/scratch/hippunfold_khanlab/ds002168_newatlas/sub-1425/surf$ wb_command -surface-information /cifs/khan/shared/data/hippunfold_models/hippunfold-atlases/tpl-multihist7/tpl-multihist7_hemi-R_space-unfold_label-hipp_midthickness.surf.gii 
Name: tpl-multihist7_hemi-R_space-unfold_label-hipp_midthickness.surf.gii
Type: Unknown
Number of Vertices: 28897
Number of Triangles: 56577
Bounds: (-39.2188, 0, -200.313, -170.313, 1.25, 1.25)
Spacing:
    Mean: 0.177764
    Std Dev: 0.030489
    Minimum: 0.156250
    Maximum: 0.220971

```

akhanf:
ok just need to get the subfield voting (and propagation) in the dag, then we are close to merging this in.. 

re: naming, I think once we're happy with the avg template and the mesh sampling, we could choose the names based on the # of vertices, e.g. 10K, 20K  and then just leave out atlas (that seems to be what they do in cortex).. I would then just keep the atlas wildcard in the subfields only..

akhanf:
ok subfield voting should be working now.. 
final edit before merging will be to set the cifti structure metadata for the template files..

akhanf:
actually, I'll leave out the setting of structure metadata and do it in another PR, not critical at this point..

akhanf:
I see the modification of run.py (importing the plugin) has broken the readthedocs -- will post an issue so we can fix in another PR..

PR #381, opened 2025-02-25
Enable the use of containers when --use-conda is not defined

akhanf:
The issue was that whenever any rule has both a conda and container directive defined, then it checks the version of conda in the container, since the behaviour will be to use the container to build the conda env. 

But for us, we just want the container to be used without conda at all, so this adds a wrapper when defining conda environments, so that it is set to `None` if the `--use-conda` flag is not set. 

Looks to be working fine in dry-run, will do a full test now..

### DIFF
--- a/hippunfold/workflow/rules/common.smk
+++ b/hippunfold/workflow/rules/common.smk
@@ -1,7 +1,24 @@
 from appdirs import AppDirs
 from snakebids.paths import bids_factory, specs
-
 from functools import partial
+
+
+def conda_env(env_name):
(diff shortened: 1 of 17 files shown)


akhanf:
actually there are also a couple container edits that were needed too, just testing those out, but if those work, make sure you use an updated container (e.g. delete your old one)..

akhanf:
ok this should be good to go now, tested with updated hippunfold-deps container, and added fix that works for older and newer snakemake versions

PR #382, opened 2025-02-26
even vertex spacing in unfolded space

jordandekraker:
NOTE: merging into `atlasgen`. I said I would ake this a separate branch, but i found the atlasgen branch easy to work off since I already had it ready to go with the multihist7. 

This first commit uses a spring model to adjust vertices in unfolded space. The target spacing is the averaging spacing from the original unfolding, and I see that all results fall within the bounds of the original unfolded space. Its fast to run 10-100 iterations, but it looks like for the multihist7 even 10000 iterations was not enough to fully smooth things out. Here is an example (bbhist left):
tmp
zooming in, we can see that some vertices are still folded over each other (especially near the middle where there are many tightly packed vertices due to gyrification):
tmp2
I think that this could be solved by having a smaller step size and more iterations, however, I wonder if there might be a faster implementation (this took around half an hour to compute).


akhanf:
Looks promising -- 
Wondering if this implementation might be more suitable: 
https://networkx.org/documentation/stable/reference/generated/networkx.drawing.layout.spring_layout.html

Uses a simulated annealing approach that might be more stable.

Would just need to build an adjacency matrix from the mesh (coudl use extract_edges() and loop), and pass along the initial positions too.

We could also pass the native vertex distances as weights to the spring layout, to enforce they stay the same relative distance.

We might also want to skip the decimation in native surfaces (eg setting decimation factor to 0) to avoid making the vertex spacing more irregular. I think (hopefully) with the updated Laplace Beltrami things should still be computationally efficient with large # of vertices..

jordandekraker:
Yeah i forgot that the vertices are not actually evenly spaced in native space. We can certainly set their target spacing accordingly, but should probably also cut the decimation (which is a bit unfortunate). 

I was also playing with a laplace smoothing function using sparse cotangent weights as input, similar to the laplace-beltrami solver. However, the cotangent weights contain a lot of extreme values, likely caused by issues like overlapping vertices, tiny folds, inverted faces, or long thin faces. I also tried a bit of preprocessing to solve this, but no luck yet. Perhaps spring_layout is more robust to these issues

akhanf:
Things also might be better behaved without decimation

jordandekraker:
Just so i understand, using https://networkx.org/documentation/stable/reference/generated/networkx.drawing.layout.spring_layout.html, G would be a sparse connectivity matrix where connected vertices have a value of:
1) 1 or 0,
2) distance or 0, or
3) cotangent weight or 0?

initially i would think 1) should be fine since we will define their positions and target distances anyway

akhanf:
> Just so i understand, using https://networkx.org/documentation/stable/reference/generated/networkx.drawing.layout.spring_layout.html, G would be a sparse connectivity matrix with values of:
> 
>     1. 1 or 0,
> 
>     2. distance or 0, or
> 
>     3. cotangent weight or 0?
> 
> 
> initially i would think 1) should be fine since we will define their positions and target distances anyway

Yes I think option 1 will work, and yes you should be able to define it with a sparse matrix, then use https://networkx.org/documentation/stable/reference/generated/networkx.convert_matrix.from_scipy_sparse_array.html to read it in to networkx (yes, would be pretty huge if not sparse!)

akhanf:
Or if we want to make use of the native vertex spacing, could use something like: e^(-ad) , where d is the vertex edge distance, would lead to lower weight as distance is increased (which should relate in ower attractive force) - here `a` would be a tunable scaling factor.. but I think ok to just try no weights first and see how it goes..

jordandekraker:
So networkx's `spring_layout` ran with method 2. for about 4 hours (max 5 iterations instead of default 50; decimation reduced to 0.05). I'd love to show the end result but it failed to save because the output was a dict instead of an array (oops)

Running again (max 5 iterations; 0.75 decimation) and saving properly gives this:
image
(the crosshair is massive because the surface was shrunken, could turn this off with scale=True)

I don't think its worthwhile to use networkx. We could use a graph-based approach to optimize my original code, or we could also just let it run for 10 000+ iterations and set correct decimation and target lengths. We'd probably be looking at 1hour runtime with no decimation though. I'll experiment a bit more with the graph methods

akhanf:
Wow that looks weird.. ya probably better off optimizing our own implementation so it can run longer. Having it use the native distances somehow might help with vertices crossing over each other..

jordandekraker:
I still haven't been able to get my laplacian graph solver working, so i rolled it back to a clean and stable version here.

The logic of my attempt was something like this:
apply cotangent_laplacian(vertices, faces) as in laplace_beltrami.py
followed by
```
def laplacian_smoothing(vertices, faces, iterations=10, lambda_factor=0.1):
    """Performs Laplacian smoothing with a uniform edge length constraint."""
    L = cotangent_laplacian(vertices, faces)
    I = sp.identity(len(vertices))

    # Compute uniform edge lengths
    edges = np.vstack({tuple(sorted(edge)) for face in faces for edge in zip(face, np.roll(face, -1))})
    target_length = np.mean([np.linalg.norm(vertices[i] - vertices[j]) for i, j in edges])

    v_new = vertices.copy()
    for _ in range(iterations):
        # Solve (I - λL)v = v_old
        A = I - lambda_factor * L
        v_new = spla.spsolve(A, v_new)

        # Adjust vertex positions to enforce uniform edge lengths
        for i, j in edges:
            midpoint = (v_new[i] + v_new[j]) / 2
            direction = (v_new[j] - v_new[i]) / np.linalg.norm(v_new[j] - v_new[i])
            v_new[i] = midpoint - 0.5 * target_length * direction
            v_new[j] = midpoint + 0.5 * target_length * direction

    return v_new
 ```
Note that the target_length would still need to be set here

One issue is that because of some bad vertices, cotangent_laplacian returns a few extreme values. Even when i try to clip or mask these, it still takes a long time to solve and starts throwing some numerical value overflow issues. I'm beginning to suspect that this sort of solution may not actually be faster than my initial attempt

jordandekraker:
i think this is a good stopping point for the unfolded distortion minimizations. I can't get any further optimizations working, and it is actually very fast to run a few hundred iterations of the spring model. The mustihit7 runs into the max iterations, ALMOST reaching the stopping condition of MAE<1e-6. This is now the slowest step for multihist7, taking 1hr for 10k iterations. 

TODO: 
- time & test whether stopping condition is reached on lower res data with fewer vertices, which would also probably make this a quite reasonable runtime
- switching unfoldreg method from greedy to ants seems to be causing T2w dryrun failures and I don't see why

I think the only other ways we could optimize the unfolded deformations would be to choose a completely different flatmapping algorithm, and I don't really want to undo so much of the work we've already done

akhanf:
just tried to give it a shot, but getting dag errors, and doesn't seem to be related to greedy vs syn for unfoldreg (ie I tried both).. Might have to wait until next week until I can look again (have some grant things to finish this weekend)..

jordandekraker:
I'll take a look at the broken DAG, just wanted to share this from the participant_create_template multihist7 results

This is curvature (applied tanh() for nicer visualization), it gives a nice sense of the unfolded shapes after 10k springy iterations
image

akhanf:
Actually was able to sort out the dag issue (note I used the naming `unfolduneven` -> `unfold`, instead of `unfold` -> `unfoldeven`, to avoid having to change the rest of the workflow) -- I'll commit that in a sec

The spring model is still breaking topology for me (albeit in small ways) -- I'm wondering if we are better off with just heavy laplacian smoothing..

akhanf:
I'm getting somewhere with the pyvista smoothing (the boundary_smoothing and feature_smoothing flags were unintuitive, and I think there are some typos in the docs), but the results with that don't seem to suffer from the issue wb_command smoothing has where everything gets pushed up against the outer edge.. running some hyperparameter tuning now to see how things look, will share soon.. 

btw, I won't be able to meet at the Tues time (conflicts with teaching) for the next few weeks..

jordandekraker:
cool, that could work. maybe also keep in mind we won't be able to set the target spacing based on the native space then.

I'll let you play around with it, let me know what you decide or if you want me to try some more

akhanf:
btw things are getting really messy with these two PRs, this one dependant on the other,  and each including many different changes (many minor/independent things that could have been separate PRs, easily merged)--  and also these branches aren't including the latest changes in dev-v2.0.0..  I am going to get the atlasgen in at least a mergeable state (even if not all objectives completed), so we can close that, along with this..  but in the future we really should try to keep the PRs focussed..

jordandekraker:
i vaguely think that we are using branches/PRs correctly. it doesn't really make sense for me to continue working on atlasgen until we solve unfoldeven. merging the latest conda into here is annoying, but not too bad

Also let me know if you're still working here on unfoldeven, otherwise i would like to 1) try a spring+laplace solution and/or 2) merge (if you are happy) and continue on atlasgen

jordandekraker:
just had a look at pyvista's laplacian smoothing. Looks quite nice, but it preserves the original boundaires. This means that it won't be "stingray" registration happening in unfolded space. I could be fine with this - its sort of a compromise between implementing minimal distortion/undersampling in unfolded space and keeping the original unfolded registration behaviour. 

I'll have a look at a hybrid approach

akhanf:
> i vaguely think that we are using branches/PRs correctly. it doesn't really make sense for me to continue working on atlasgen until we solve unfoldeven. merging the latest conda into here is annoying, but not too bad
> 
> Also let me know if you're still working here on unfoldeven, otherwise i would like to 1) try a spring+laplace solution and/or 2) merge (if you are happy) and continue on atlasgen

unfold spacing is completely independent of the atlasgen approach though, so this could just merge in before..

jordandekraker:
> > i vaguely think that we are using branches/PRs correctly. it doesn't really make sense for me to continue working on atlasgen until we solve unfoldeven. merging the latest conda into here is annoying, but not too bad
> > Also let me know if you're still working here on unfoldeven, otherwise i would like to 1) try a spring+laplace solution and/or 2) merge (if you are happy) and continue on atlasgen
> 
> unfold spacing is completely independent of the atlasgen approach though, so this could just merge in before..

I'm not going to try to code the atlas generation with no workable data to inspect along the way

akhanf:
> just had a look at pyvista's laplacian smoothing. Looks quite nice, but it preserves the original boundaires. This means that it won't be "stingray" registration happening in unfolded space. I could be fine with this - its sort of a compromise between implementing minimal distortion/undersampling in unfolded space and keeping the original unfolded registration behaviour.
> 
> I'll have a look at a hybrid approach

I did a deep dive over the weekend (and did some hyperparameter grid searchs) --  you can get it to smooth the boundaries too, but you just have to set boundary_smooth to True, and then also set the edge_angle high enough (e.g. 180 will do it).. BUT, the results it gives, end up tending to a disk, and although the spacing near the edges is not compressed, it seems to keep the interior features compressed.. I did also try workbench smoothing followed by pyvista smoothing, but still didn't give the desired result.. 

The other thing I just noticed (when running things on the multihist7 data) is the smoothing can exaggerate artifacts in the unfolding, eg when a cluster of vertices is near the boundary they seem to stay there (and you end up with a sharp ridge jutting out.. 

One more thing I wanted to try was to revisit the force-directed graph layout algorithms, but this time use sfdp, which is meant to be efficient for very large graphs.. another option might be to try to enhance the existing spring implementation to penalize edge crossing, but that might be complicated..

akhanf:
> > > i vaguely think that we are using branches/PRs correctly. it doesn't really make sense for me to continue working on atlasgen until we solve unfoldeven. merging the latest conda into here is annoying, but not too bad
> > > Also let me know if you're still working here on unfoldeven, otherwise i would like to 1) try a spring+laplace solution and/or 2) merge (if you are happy) and continue on atlasgen
> > 
> > 
> > unfold spacing is completely independent of the atlasgen approach though, so this could just merge in before..
> 
> I'm not going to try to code the atlas generation with no workable data to inspect along the way

I'm not suggesting that, just suggesting this could merge into dev-v2.0.0 instead of atlasgen (atlasgen could merge upstream changes as usual to then work on it).

jordandekraker:
Wow, this was a really neat deep dive. I think that the topology behind knitting is too smart for me to understand haha. Maybe some things can only be understood by grandmothers

Let me know if ImPrEd works out, but otherwise springs+laplace should be OK. A few minor issues to iron out in that method:
- agree that the unfolded surface pv.clean() is important to remove overlapping vertices, and we should maybe take the index of bad vertices back to the folded surface too. 
- in this example, did you apply springs and then laplacian smooth? I played around a bit with doing them interleaved, but its a bit annoying to work aout a good ratio between them, so applying laplacian at the end might be nicer
- are we concerned that decimation will remove vertices preferrentially from low curvature areas? I believe this is the behaviour of decimate_pro() (e.g. fig 7,8 from its [reference paper](https://dl.acm.org/doi/pdf/10.1145/133994.134010)). I think that just decimate() is actually what we want (also seen [here](https://docs.pyvista.org/examples/01-filter/decimate))

akhanf:
Yeah, was just 5 iterations of strength 1.0 smoothing after the spring smoothing (will try to clean up and commit later today if I get a chance), I found doing more would lead to compression at the edges, though I did not evaluate different smoothing strengths.. maybe if we decide this is the way to go, we can do some hyperparameter tuning to assess size distortion (as defined by differences in edge lengths native to unfold, and surface areas). 

I think decimation removing low curvature areas is fine, as long as the edge lengths are mostly preserved going to unfolded space, so that the 2d space still has correct physical spacing.

jordandekraker:
Sounds good.

I still think we should use decimate() instead of decimate_pro() though since part of the notion of "stingray" registration was to ensure even sampling in unfolded space. A flat subiculum could lead to just a few thickness or T1w/T2w intensities being used in unfodled registration

akhanf:
ok, some interesting news -- turns out tulip supports a wide range of graph layout algorithms, including a whole bunch that are implemented in this Open Graph Drawing Framework: https://ogdf.uos.de/. A popular and fast one that people seem to be using alot is FMMM https://ogdf.netlify.app/classogdf_1_1_f_m_m_m_layout#details

It works really well, and is extremely fast. What is particularly unsettling is that I can just pass it my native mesh, and I get basically the same unfolded output!

image

Anyways, thats something to think about -- but for now I'm going to try to incorporate it from our unfolded surface (one can set desired edge lengths), and ensure points are scaled appropriately, will keep you posted..

jordandekraker:
neat, sounds like it'll work

And yeah i'm vaguely aware of lots of other conformal mapping methods. The laplace coodinate system is nice in that it keeps anatomically-motivated edges, and i slightly worry that with too drastic of changes in unfolded space we may see things fail to align fixed edges - that is, MTLc borders could be warped over to HATA borders. Perhaps we could use our masks of the boundary vertices as additional modalities in the unfolded registration to avoid this

akhanf:
Actually, it looks like it always does weird things with scaling and orienation of the resulting nodes, which is going to be a pain to deal with, so I think just sticking to the current approach  of spring + minor smoothing will be preferred.. I'll make clean up and make those commits shortly.. 

By the way the conformal mapping approaches typically work differently (and usually preserve angles by definition of being conforming), which is something we could consider in the future (eg solving laplace jointly.. ), these graph layout algorithms are another beast altogether..

akhanf:
btw since this branch was based off the atlasgen one, it now includes all those commits too..  what I can do is rebase this one so it leaves those out -- but that can mess things up if you then make additional changes, so I'll hold off until we are ready to merge perhaps

I think now is maybe a good time to merge this in (once I get tests passing), as we can make any additional tweaks, like surface cleaning in unfold+native,  in other small PRs.

jordandekraker:
i don't expect it will take me long to finish atlasgen with this now in place. if we merge into that then i can probably merge that one tomorrow

akhanf:
ok lets do that. I'll merge this into atlasgen when the tests pass

akhanf:
before I do that though, I'll bring atlasgen up to date with dev-v2.0.0 (otherwise this merge is a big one)..

PR #388, opened 2025-03-12
Add functionality for hippunfold-singlesubject

mackenziesnyder:
- WIP still because I have not been able to test this end-to-end due to constraints with cbs basic
- command to run on a single subject non-bids structured file: 
`hippunfold-singlesubject --input /path/to/input.nii.gz --output /path/to/my/output_folder --subject 001 --modality T1w `
- add: `--use-singularity` or `--use-conda`
- optional, add: `--temp-dir /path/to/tempdir`

### DIFF
--- a/hippunfold/workflow/scripts/single_subject.py
+++ b/hippunfold/workflow/scripts/single_subject.py
@@ -1,0 +1,111 @@
+from argparse import ArgumentParser
+import os
+import tempfile
+import shutil
+import subprocess
+
+def gen_parser():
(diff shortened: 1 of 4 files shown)


hippunfold/workflow/scripts/single_subject.py
akhanf, line 1: Nice! I think you could move this script to be next to the run.py -- 
I also think we could rename it something pithier -- 
hippunfold-quick
hippunfold-ez
hippunfold-easy
hippunfold-now

thoughts? 

the script could be called run_<suffix>.py

Issue #387, opened 2025-03-11
inverted myelin mapping


jordandekraker:
I just checked and the documentation for c3d's `-divide` doesn't match the behaviour (of our current version at least). This means that myelin maps are currently inverted.

akhanf:
closed with #392

PR #392, opened 2025-03-12
Invertedmyelin

jordandekraker:
a very quick fix for https://github.com/khanlab/hippunfold/issues/387

Note: I'm not sure if this is the case for all c3d versions! this was run from autotop_deps.sif


jordandekraker:
@Dhananjhay to pin c3d version, then we will merge

PR #404, opened 2025-03-20
no user site in python path

jordandekraker:
https://github.com/khanlab/hippunfold/issues/400

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,7 @@ COPY . /src/
 
 # avoid pre-downloading the models to make for lighter container
 # ENV HIPPUNFOLD_CACHE_DIR=/opt/hippunfold_cache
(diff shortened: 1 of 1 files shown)


akhanf:
looks fine, but I'll trigger a test build from the branch so we can try it out first..

akhanf:
ok the container is building now -- perhaps the user can try it out when ready?
https://github.com/khanlab/hippunfold/actions/runs/13972470199

jordandekraker:
I tried to test this out but have only python3.10 (3.9 in the container) so its hard for me to tell whether --user installed libraries are truly being ignored (they seem to also be ignored in the previous container but i think that's just because of the different ypthon version).

akhanf (approved):

Issue #403, opened 2025-03-20
limited range in dentate coords-AP


jordandekraker:
The AP coords in the dentate seem a bit odd. The middle range spans most of the hippocampus, while the upper and lower quartiles are mostly just in the tip of the uncus and tail. 

Here is an example of the hipp AP coords:
![Image](https://github.com/user-attachments/assets/e590fade-c56c-48d0-a32e-18f056490a28)
And dentate (same subject):
![Image](https://github.com/user-attachments/assets/f3fe9112-e96a-4d91-bdcd-98d7324490ea)

I've checked and the boundary conditions are correct. 
This is the cause of the violin-like shape of the unfolded dentate. 

The old (volumetric) laplace solver solved AP over the dentate and hipp jointly. What we're doing now is truly a slightly different coordinate system. Now the AP coords need to enter into the concave portion of the anterior hippocampus before reaching the hippocmapal body, taking away a "shortcut" that the AP coords could previously take trough the inner elbow of the uncus. In addition to having to travel further, I think that the laplace solution we're getting could actually be correct - that is - the widening of the dentate in the body of the hippocampus may truly slow the spread of the equipotential differences (like how a wide river flows slowly). 

In either case, it seems a bit odd to me to not have the dentate AP coords matched to nearby hipp AP coords (as in a joint solution). We could consider "stitching" the dentate and hipp surfaces (proximal edge to distal edge) to solve jointly. This will require some additional rules though. 

 What do you think, @akhanf?

akhanf:
So looks like there are two separate issues -- 1) laplace solution slowing down in wide areas (which I agree is the correct laplace solution), but perhaps this solution is not a desirable one for defining coordinates, and 2) hipp and dentate A-P being out of sync (which is exacerbated by issue 1).. 

Stitching the dentate and hipp seems like it would be challenging and we would have a number of topology issues we may need to solve, and ultimately would be pretty hacky.. 

I am thinking it would be better to resolve 1 by modifying what we solve -- e.g. if we use a solution that provides geodesic distances instead (e.g. like fast marching), then this could also lead to a more uniform unfolded space..  

We could then try to deal with hipp to dentate A-P alignment in a different way (if any further correction is needed).. 

Have just started playing around with this (thanks GPT for implementations!) and a mesh-based version of the scikit-fmm approach seems to work quite well!

e.g. laplace:

[...]

PR #406, opened 2025-03-21
Fast marching

akhanf:
This adds fast marching to improve AP coordinate calculations (see #403 )
Implemented in the same laplace_beltrami script, and enabled through the params when dir=AP.


akhanf:
update: fast marching might be introducing distortions in the unfolded surfaces -- investigating, but likely will not merge this PR in..

akhanf:
ok with the other issue resolved, it does still look like fast marching is distorting things. Below is a head to head comparison with/without fast marching. Note that this is with and without for both building the template and when applying it to a subject (bbhist). The surfaces shown are 7K and 2K surfaces for hipp and dentate. 

Without fast marching:
image

With fast marching:
image

So I think it's safe to just close this PR without merging..

jordandekraker:
was thinking more about this and had a lovely little chat with chatgpt: https://chatgpt.com/share/67f68518-a544-8009-b11e-038a0c372911

instead of averaging forward and backward passes, combining them like this basically does exactly what we want. Here's an example of the unfolded dg surface:
dg

This is very fast and makes me wonder if we should throw out L-B altogether and use this instead!

akhanf:
I'm still getting the surface resampling segfaults when I run this? Doesn't happen on the dev-v2.0.0 branch so something specific to this PR but haven't looked closely at it yet.

jordandekraker:
sorry had some unpushed work. reflective bounds stops added noise exceeding 1.0

jordandekraker:
also note that i'm switching `decimate_pro` here to `decimate`. I don't think that this was actually causing the overlapping vertices, but i do generally think its a good idea to avoid undersampling the relatively flat subiculum

jordandekraker:
Had a look at the results and i'm pretty happy with the multihist7 using relative geodesic distances. i did notice some wrinkles in the unfolded surfaces, possibly due to geodesic "shortcuts" but there were really not bad on the multihist7 and should be even more mild on smoother samples. The geodesic method should also generally lead to stronger gradients in the middle of the hippocampus, and therefore more vertices in the middle, which i see as an advantage. Maybe we should leave in the LP-BT solver for now but could consider cutting it entirely in the future

akhanf:
ok great I'm giving it a whirl right now on multihist7, and  I'll try to have a look at it later tonight as well on a standard image so we can hopefully decide tomorrow.

akhanf:
I'm still seeing the quality issues for the resampled surfaces. Here is bbhist hemi-R from the dev-v2.0.0 branch (no fast marching) vs with fast marching. Did you look at the resampled surfaces in your cases, or were you just inspecting the unfolded space?

image
image
image

jordandekraker:
hmmm yeah initially i thought this was just because of the downsampling, but i agree it seems to be disproportionately stretched in sulci while still good quality on the gyri, which is probably because of the "wrinkles" in unfolded space. it doesn't look quite so bad in the other samples, but we might still be able to do better. 

I have one more idea: will take the LPBT solution and rescale the values to make a uniform distribution. I'll make a separate PR for this.

PR #407, opened 2025-03-23
Remove work folder, use temp() instead

akhanf:
Taking advantage of the fact that we have very few open workflow PRs to do some of the clean-up PRs. 

Use root for everything, and use temp(...) for any outputs that were put in work.

Anything that was in work will now just be put in the main output location, and deleted when no longer needed, unless `--notemp` is used.

This makes the workflow simpler and less error prone since we don't have to remember which files are in work and which are in root (all are in root), and also much easier to set a file as temporary, since we now only need to specify it when it is declared as an output, by putting `temp(...)` around it. 

Note: I haven't changed which files are put into work (now temp())  in this PR, just kept things as they were, can do that in further clean-up/output-refactoring PRs..

### DIFF
--- a/hippunfold/config/snakebids.yml
+++ b/hippunfold/config/snakebids.yml
@@ -287,14 +287,6 @@ parse_args:
     action: "store_false"
     default: True
 
-  --workdir:
-    help: |
-      Folder for storing working files. If not specified, will be in "work/" subfolder 
-      in the output folder. You can also use environment variables when setting the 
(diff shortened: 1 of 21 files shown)


Dhananjhay (approved):
I successfully ran the end-to-end pipeline using the lowresMRI dataset twice—once without `--notemp` and once with it. In both cases, there was no work folder in the output.

akhanf:
Great, I'll merge this in now!

PR #409, opened 2025-03-24
unfold atlas meshes at different densities

akhanf:
- ~currently densities are named as 100x (100%) or 50x (50% initial size), implemented by simply resampling input ref 2d image~
- minor refactoring (renaming rule/script)

TODO: 
  - [x] rename based on # of vertices (e.g. 32k)


akhanf:
OK phew I think this is ready to merge. 

The CLI for densities not fully completed yet but it is functional, will post an issue for the remaining bits..

akhanf:
Merging this in ahead of brainhack..

PR #412, opened 2025-03-26
Separate surface functions into a module

akhanf:
@Dhananjhay I know you started something like this but in another repo - this is more of a direct port of existing functions, mainly to try to avoid the redundancy in surface functions across scripts. 

I started this, but got distracted with #409 and didn't finish it, but it is almost there if anyone is able to pick this up eg for brainhack?


Dhananjhay:
Yes, I can takeover!

PR #414, opened 2025-03-27
removed unfold spring and unfold smooth

jordandekraker:
This is just an experiment to see what will happen, and discuss. This effectively throws away the "stingray" unfolded registration and sticks to the original 2:1 aspect ratio rectangle. It still uses the workflow to generate new atlases though, in this case still using the multihist7

image

The original goal of the "stringray" space was to have more even vertex sampling, minimizing distortion between folded and unfolded space. We also generally wanted better quality reference surfaces, which we have achieved whether we include stingrays or not. However, there are also some pros to skipping this:

Pros:
- faster and (slightly) simpler workflow
- could drop the padding from unfolded space registration
- can avoid errors arising from springs + smoothing (e.g. edge vertices being crammed together)
- easier mapping back to old surface densities (i.e. by interpolation in unfolded space) (provided we use the same laplace coords and not geodesic distances)

To be honest, i think that we could drop the springs+smoothing, since the vertex spacing is not terribly different here (comparable to stingrays and maybe better for the edges). Plus this makes things easier to map onto HippoMaps and visualize with the folded+unfolded space.

One final thing to consider is that in unfolded template generation, we could drop affine and keep only syn registration (since all boundary conditions are initialized in the same place. I will try this out now, but i don't expect it to make a difference.

### DIFF
--- a/hippunfold/config/snakebids.yml
+++ b/hippunfold/config/snakebids.yml
@@ -616,7 +616,7 @@ laminar_coords_res:
   dentate: 0.1x0.1x0.1mm
 
 resample_dseg_for_templatereg: 0.3x0.3x0.3mm
-unfoldreg_padding: 64x64x0vox
+unfoldreg_padding: 0x0x0vox
 
 density: #temporary var
   hipp: 
--- a/hippunfold/plugins/atlas.py
+++ b/hippunfold/plugins/atlas.py
@@ -18,7 +18,7 @@ import json
 import os
 
 # Global variable to store the commit hash
-ATLAS_REPO_COMMIT = "7c292113c669a629773a644f82485ffde1842f52"
+ATLAS_REPO_COMMIT = "c1a53ecade939ead9de8f9169c6a4ddff0c73c3d"
 
 
 def get_download_dir():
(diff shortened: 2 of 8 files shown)


jordandekraker:
one additional though: with the meshgrid reference mesh generation and no "stingrays", we can always reshape vertices into an unfolded 2:1 aspect ratio space. There will never be a number of vertices that don't fit this ratio. This is very nice for later analyses, like plotting with 3D surfaces or 2D images (plt.imagesc), and for doing things like 2d convolution operations on meshgrid surfaces (not necessarily recommended since the real-world spacing will be uneven, but still a cool way to do rough exploration)

jordandekraker:
another downside though: we are sticking closer to the previous measure of gyrification with this. The closeness to average native surface spacing in stingrays may have actually been better for this (but that also was still not quite perfect!). Perhaps it should be computed as a ratio with a "skeleton" mesh with smoothing but preserving the topology and edge vertices

akhanf:
I think the goal of getting native-vertex-spacing in the unfolded space is still a good one, but we aren't very close to that currently, so I am happy to just resort to the simpler solution of not performing those steps, especially since the performance seems comparable..

akhanf:
We could still explore different gyrification measures that make use of e.g. smoothing in the unfold and/or native space, even if we aren't using smoothed surfaces for coords..

akhanf:
made a few more fixes -- re: gyrification, we just discussed how we could apply the prev spring+smooth steps, but only for that metric.. I think that's the last thing to update here --

akhanf:
OK I've made things alot simpler since it looked like using the same density surfaces for hipp and dentate seemed reasonable enough.. Note this relates to a different resample rate (2x more for dentate). 
Also note that with the unfoldrect approach, the mapping to density could be just determined using the unfold dimensions, but since I've already done it with the checkpointing approach we'll just keep it in (in case we ever use something other than unfoldrect in the future).. 

This should be ready to merge now, I've also updated hippunfold-atlases.

PR #415, opened 2025-03-27
Stitchlaplace

jordandekraker:
this is the stitching solution outlined for https://github.com/khanlab/hippunfold/issues/403

TODO: check for off-by-1 errors  in indexing (haven't tested this yet!)


jordandekraker:
I believe I sorted out the issues here, but I'm still getting approximately the same results, with the dentate showing the greatest coordinate changes in the tip of the uncus and tail. Will experiment a bit more with this in the next couple days

jordandekraker:
This is close-ish to working, but i seems to be introducing some jumps in the AP coords. This messes up the unfolding, example here:

tmp

weirdly, these kinds of issues are not only at the edges (even though stitching was only performed at the edges), and so I think the stitching may actually be causing more global disruptions to the L-B solver.

PR #422, opened 2025-04-01
Cache conda envs

Dhananjhay:
By default, Conda environments are cached and downloaded in the .snakemake directory relative to the invocation directory. This means that each time a user runs the pipeline with a different output directory, the Conda environments are re-downloaded. This PR introduces a centralized cache repository for Conda environments, similar to how Singularity caches images in the .singularity-cache directory. Users who prefer a custom location for storing Conda environments can still override this by specifying the `--conda-prefix` flag when running HippUnfold.

### DIFF
--- a/hippunfold/plugins/atlas.py
+++ b/hippunfold/plugins/atlas.py
@@ -16,19 +16,10 @@ logger = logging.getLogger(__name__)
 
 import json
 import os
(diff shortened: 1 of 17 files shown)


akhanf:
made a couple updates, but looks good to me otherwise (note: haven't actually run it yet)

akhanf (approved):

PR #429, opened 2025-04-15
LP uniform

jordandekraker:
This rescales the Laplace coords to have a uniform distribution (while leaving source/sink alone at 0 or 1). This also makes the gradient steeper in the middle, helping provide more aligned vertices in those areas which need it most. Unlike the fast marching method (https://github.com/khanlab/hippunfold/pull/406), this also ensures no "wrinkles" taken by geodesic shortcuts. It doesn't guarantee alignment between the nearest hipp and dentate vertices like (https://github.com/khanlab/hippunfold/pull/415)

### DIFF
--- a/hippunfold/workflow/scripts/laplace_beltrami.py
+++ b/hippunfold/workflow/scripts/laplace_beltrami.py
@@ -140,9 +140,13 @@ boundary_conditions = dict(
 coords = solve_laplace_beltrami_open_mesh(vertices, faces, boundary_conditions)
 
 # make fully uniform distribution
-domain = np.setdiff1d(np.arange(len(coords)), np.concatenate([src_indices, sink_indices])) # ignore src/sink indices
+domain = np.setdiff1d(
+    np.arange(len(coords)), np.concatenate([src_indices, sink_indices])
+)  # ignore src/sink indices
 sorted_indices = np.argsort(coords[domain])
-uniform_values = np.linspace(0, 1, len(domain)+2)[1:-1] # ignore first and last values
+uniform_values = np.linspace(0, 1, len(domain) + 2)[
+    1:-1
+]  # ignore first and last values
 coords[domain[sorted_indices]] = uniform_values
 
 # save the coordinates to a gifti file
--- a/hippunfold/workflow/scripts/laplace_beltrami.py
+++ b/hippunfold/workflow/scripts/laplace_beltrami.py
@@ -139,6 +139,13 @@ boundary_conditions = dict(
 
 coords = solve_laplace_beltrami_open_mesh(vertices, faces, boundary_conditions)
 
+# make fully uniform distribution
+domain = np.setdiff1d(np.arange(len(coords)), np.concatenate([src_indices, sink_indices])) # ignore src/sink indices
+sorted_indices = np.argsort(coords[domain])
+uniform_values = np.linspace(0, 1, len(domain)+2)[1:-1] # ignore first and last values
+coords[domain[sorted_indices]] = uniform_values
+
+# save the coordinates to a gifti file
 data_array = nib.gifti.GiftiDataArray(data=coords.astype(np.float32))
 image = nib.gifti.GiftiImage()
 


jordandekraker:
the unfolded and resampled corobl surfaces look immaculate, and this is my preferred solution of all the ones we've tried for https://github.com/khanlab/hippunfold/issues/403. 

One thing to generally consider is that the AP coords now scale with how PD-wide a given subject's surface is (and vice-versa). I think that this is generally in-keeping with the idea of minimizing distortion in unfolded space based on a given subject's actual surface size though, plus it will additionally undergo unfolded alignment. So I feel that this is an upgrade to the initial Laplace coordinate framework across the board.

akhanf:
Are you able to post any screenshots? Won't be able to give it a try until tomorrow..
________________________________

jordandekraker:
just to clarify a bit how i got here:

Originally after chatting with chatgpt we were going to do probability integral transform, which involves rescaling according to cumsum (or integral). then i thought about including surface area as well, but then i realized i would have to do that in 2D (otherwise long+skinny triangles would have to opposite effect of short+fat triangles).  i also wanted to avoid binning. 

Then i thought more about whether a simpler 1D case would work, opened a new session, and got this as the simplest way to turn a non-uniform distribution into a uniform one. I was skeptical at first but thinking more and testing it out, it seems pretty solid.

akhanf:
I like the general idea, especially the simplicity of it. However, the distortion that happens as the shape tapers out in the tail is quite bad compared to the laplace resampling though, but I think it could be corrected in a fairly straightforward way (perhaps already the surface area idea you were thinking about)..

The intuition is that if we had a rectangular surface, then the histogram we want to remap to is a uniform one. But if its e.g. now tapered on one end to become like a triangle, then the histogram we want to remap to is a modified uniform distribution, ie it is distorted by the surface area in each bin, so it is now a linear distribution.  The approach you're taking is one used for exact histogram matching, in this case matching to a uniform histogram, but you could use the same idea (ie sorting) but use any other target histogram. So once you have estimates of surface area for each "bin", then you can use that to make your target histogram and transform the data.

jordandekraker:
It sounds like what you're suggesting is something mid-way between the original (laplacian) histogram which respects surface areas and a uniform histogram. An even simpler way to compute this could be:
```
coords = w*coords + (1-w)*uniform_coords
```
, where w is the weighting of the original vs. uniform (e.g. 0.5). 

I actually don't mind that the tail and uncus are slightly underrepresented using the uniform distribution. It had previously bothered me that they're so prominent in unfolded space despite having such small surface area (and also typically noisier associated data due to issues like partial voluming, poor segmentation, etc). For that reason I had made an optional post-hoc way to minimize this in the unfolded space in hippomaps ([here](https://github.com/jordandekraker/hippomaps/blob/52822a01cfe64067babc11e76475141e76a4459d/hippomaps/plotting.py#L309)), which similarly relies on the integral of surface areas. baking this shift into the coordinate framework a-priori actually seems like an advantage to me. This will also mean that unfolded registration is driven more by data from the head&body of the hippocampus, where quantitative values tend to be a bit more trustworthy.

So while i'm open to doing some midway histogram, i actually think that the uniform distribution provides a more desirable representation in unfolded space

akhanf:
Not quite was I'm suggesting, I'm suggesting basically just changing the target distribution from uniform to something that is still very smooth but not uniform, and still using the exact cdf matching approach you're using here. In a meeting right now but should be able to try something shortly after.

I'm thinking probably don't even need to use the surface areas directly, can just take the histogram of scalar values and do some heavy kernel smoothing to get the target distribution.
________________________________

jordandekraker:
Thinking more about this, I believe that the integral of surface areas would generate a transform which we could use to move from original to a uniform distribution of coords, whereas the sort method goes straight to a uniform distribution. Both methods should never reorder (or break sort indices), nor exceed the 0-1 bounds. I also think the the difference `coords_orig - coords_uniform` should be proportional to the integral of surface areas (this would only be off because of issues like binning or anisotropic triangles). 

So i believe what we're doing with this little trick is skipping the computation of a transform (which would otherwise require some more thought to work out elegantly). And by removing the impact of surface area on laplace fields, we're actually computing now what i originally thought the laplace was computing - a field in which the gradient along the A-P path is constant instead of one where the gradient diverges in direction to meet wider surface areas while the gradient magnitude remains constant. I think this might be called an Eikonal field. 

Just now seeing your comment about adjusting the histogram by smoothing. I believe smoothing would still be a way to meet midway between the actual laplace histogram and a uniform one. Binning, discrete integrals, and smoothing would all be ways to approximate that transform (or midway towards it). But in the end, the fully uniform histogram is giving the desired behaviour and I don't really want to add more vertices to the tail and uncus (AKA increasing surface area of the tail&uncus in unfolded space)

jordandekraker:
reading more, a normal way to solve an Eikonal field is with two-pass fast marching as we did in https://github.com/khanlab/hippunfold/pull/406. However, Eikonal fields aren't well posed partial-differential equation, and this is only an approximation. Interestingly, the solution we have now might actually be a true Eikonal solution...

akhanf:
> Thinking more about this, I believe that the integral of surface areas would generate a transform which we could use to move from original to a uniform distribution of coords, whereas the sort method goes straight to a uniform distribution. Both methods should never reorder (or break sort indices), nor exceed the 0-1 bounds. I also think the the difference `coords_orig - coords_uniform` should be proportional to the integral of surface areas (this would only be off because of issues like binning or anisotropic triangles).
> 
> So i believe what we're doing with this little trick is skipping the computation of a transform (which would otherwise require some more thought to work out elegantly). And by removing the impact of surface area on laplace fields, we're actually computing now what i originally thought the laplace was computing - a field in which the gradient along the A-P path is constant instead of one where the gradient diverges in direction to meet wider surface areas while the gradient magnitude remains constant. I think this might be called an Eikonal field.
> 
> Just now seeing your comment about adjusting the histogram by smoothing. I believe smoothing would still be a way to meet midway between the actual laplace histogram and a uniform one. Binning, discrete integrals, and smoothing would all be ways to approximate that transform (or midway towards it). But in the end, the fully uniform histogram is giving the desired behaviour and I don't really want to add more vertices to the tail and uncus (AKA increasing surface area of the tail&uncus in unfolded space)

I think we're not talking about the same thing here, anyways hopefully can clarify in the call. Ie the correction I'm talking about still maintains the sorted order as the perfectly uniform distribution does, and taking the kernel to be very large you end up with the uniform distribution.. 

Eikonal equaitons and fast marching is going to give a very different scalar field that we found doesn't really work well for unfolding and resampling, so I am skeptical about going back to that approach in #406

jordandekraker:
yup, sounds good. And I agree that the ideal distribution is not uniform either, its just hard to decide what it should be when our current AP coords are distorted (especially in the DG). This uniform distribution overrepresents the middle w.r.t. geodesic distances (though i'm still not sure that it overrepresents this in terms of actual surface areas which *should* be smaller in unfolded space at the edges). Let's merge for now and consider optimizing later on

PR #431, opened 2025-04-15
Use UV

Dhananjhay:
This PR switches out poetry for uv for managing packages and makes appropriate changes to the Continuous integration (CI) pipeline for quality checks and dry runs.


akhanf:
Nice! I'll try to give it a try soon. 

This reminds me it would be nice to make some lab/superlab repo templates with uv and related ci, or at least update snakebids create to provide a uv option.

Dhananjhay:
Agreed! Right now in this PR hippunfold's CI pipeline uses locally added action.yml based on uv package manager for running python tests through python-testing.yml but we can add the same action.yml in https://github.com/khanlab/actions which then users can then pull from and use it in their own repos. I can also get started on adding uv option for command line command `snakebids create`.

Dhananjhay:
With the recent discussion re: package manager it no longer makes sense to use UV.

PR #435, opened 2025-04-16
Atlas cleanup

jordandekraker:
The first few commits here are necessary minor bug fixes, but the latter few change the way atlases are generated:

- the same resampling is applied to hipp and dentate surfaces
- the mapping between `resample` and `density` is defined upfront in atlas.py instead of on-the-fly

Therefore, the user should consider `ATLAS_DENSITY_CHOICES` and/or `RESAMPLING_FACTORS` as initial guesses, and reconsider after viewing the results from `resample_to_density_mapping`. 

Ideally we could create a prompt at this checkpoint the confirm that these two are appropriate, giving the option to redefine one/both before proceeding. I'm not sure how to create such a prompt, so will leave it like this for now. This should be safer, since the previous naming scheme will not generate the same density names between hipp and dentate, and its also possible that different resampling factors could map to the same density names.


akhanf:
So the current dev-v2.0.0 code is a bit of a relic from the earlier code, which had to deal with potentially each subject stingray having different densities (which is why we needed the checkpoint to create the unfold surface first, and to perform the mapping to density in ##k naming). But in general the idea was a wide range of resample factors are used which generate multiple templates, and then the user would need to intervene to actually look at the data and decide what densities to use. 

But now with unfoldiso surfaces, we can actually know ahead of time, what resampling factor will lead to what density, so if we want to we could just have the densities as inputs instead of resampling factors, and back-calculate the resampling factor that would produce it.. 

Instead of doing that, the current code just dropped all the resample factors except for those that produce the densities that we desired.

jordandekraker:
ah yeah true, there will be a 1:1 relationship between resample and density since they are defined only in unfolded and not native surfaces

I won't be able to get to this tonight, but will try to clean that up tomorrow (unless you get to it first)

jordandekraker:
I generated an atlas with the densities described here; this can now be used in dry-run tests

I'm still open to different densities, the current have spacings of ~1mm, ~0.5mm, ~0.33mm

akhanf:
Sounds good. 

Made a couple minor somewhat related updates here to enable pybids parsing of the dseg files (the trick here is to use the --derivatives flag), and a fix for the git checkout of latest atlases (wasn't updating correctly)..
Also I did put the multihist7 on datalad now, so now with both updates the instructions for building a new atlas is as simple as:

```
datalad install https://gin.g-node.org/hippunfold/multihist7
datalad get -d multihist7 multihist7/sub-*/anat/*dseg.nii.gz
hippunfold  multihist7 out_multihist7 group_create_atlas --modality dsegtissue --derivatives multihist7 --new-atlas-name multihist7 --use-conda 
```
(then would still need to commit the new atlas from hippunfold-atlases)

akhanf:
just looping back to this after my end-end test -- will have a couple fixes to make. I'll also probably remove the checkpoint job (with the current edits its no longer being used as a checkpoint and using global config anyways)...

akhanf:
OK this looks good to merge, my fixes don't change the behaviour (except dropping native surfs from generated atlas), and mainly deal with CLI fixes and removing the checkpoint.

PR #436, opened 2025-04-17
fix for unpro

akhanf:
reduction -> target_reduction


jordandekraker:
good catch! sorry about the error

Issue #434, opened 2025-04-16
Warning needed for running two hippunfold processes at the same time


mackenziesnyder:

ajschadler12:
Quick tip: if running multiple subjects in parallel with separate hippunfold processes, you can sidestep this issue with isolated output directories.

You'll end up with something like:

```
hippunfold_work/
    sub-abcd/
        .snakemake/
        config/
        logs/
        sub-abcd/
        .snakebids
```

You can move `<hippunfold_work>/sub-abcd/sub-abcd/` to a final output directory.

E.g. something like:

```sh
BIDSDIR="/home/user/bidsdir"
HIPP_WORK="${BIDSDIR}/derivatives/hippunfold_work"
HIPP_OUT="${BIDSDIR}/derivatives/hippunfold/"

mkdir -p $HIPP_WORK
mkdir -p $HIPP_OUT
for sub in 1 2 3 4 5
do
    sub_out="${HIPP_WORK}/sub-${sub}"
    mkdir -p "$sub_out"
    hippunfold "$BIDSDIR" "$sub_out" # rest of options ...
    mv "${sub_out}/sub-${sub}" "$HIPP_OUT"
done
```

PR #437, opened 2025-04-17
Mackenzie/same output folder warning

mackenziesnyder:
- checks if a config/snakebids.yml file already exists in the output directory specified by the user
- if it does, a warning is sent to the terminal about two possible snakebids processes writing to the same output folder at the same time. Only a warning as if a previous process has already been completed, it should be ok if the config file is overwritten by the second process.
- for issue: #434 
- please let me know if this fully covers the problem discussed during the meeting on April 16th - I created the pr based off what I had noted down in the attached issue

### DIFF
--- a/hippunfold/run.py
+++ b/hippunfold/run.py
@@ -1,6 +1,8 @@
 #!/usr/bin/env python3
 import os
 from pathlib import Path
+import sys
+import warnings 
 
 from snakebids import bidsapp, plugins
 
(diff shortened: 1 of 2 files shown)


akhanf:
sorry wasn't able to comment on your issue in time -- we still want to allow users to write to the same output folder, the issue is when multiple runs are initiated at the same (or close to same) time.. 

e.g.  somethign like:

```
for sub in 1 2 3; 
do
    hippunfold_quick -i sub-1.nii.gz -o output -m T1w
done

```
or 
```
hippunfold_quick -i sub-1.nii.gz -o output -m T1w & 
hippunfold_quick -i sub-2.nii.gz -o output -m T1w & 
hippunfold_quick -i sub-3.nii.gz -o output -m T1w
```
So the changes made in the PR so far aren't really desired.. I think we can just leave this for now, perhaps just make a FAQ note about it ..

mackenziesnyder:
> sorry wasn't able to comment on your issue in time -- we still want to allow users to write to the same output folder, the issue is when multiple runs are initiated at the same (or close to same) time..
> 
> e.g. somethign like:
> 
> ```
> for sub in 1 2 3; 
> do
>     hippunfold_quick -i sub-1.nii.gz -o output -m T1w
> done
> ```
> 
> or
> 
> ```
> hippunfold_quick -i sub-1.nii.gz -o output -m T1w & 
> hippunfold_quick -i sub-2.nii.gz -o output -m T1w & 
> hippunfold_quick -i sub-3.nii.gz -o output -m T1w
> ```
> 
> So the changes made in the PR so far aren't really desired.. I think we can just leave this for now, perhaps just make a FAQ note about it ..

Ahhh ok I understand now - thanks!

Issue #443, opened 2025-04-22
Output densities for dev-2.0.0


akhanf:
With #435 is merged in, we now have ~0p3 (18k), ~0p5mm (8k) and ~1mm (2k) surfaces.

However, our previous set of surfaces were 0p5mm (14,266 verts), 1mm (3,870 verts), and 2mm (774 verts).  Since we are now using an unfoldiso grid we can't really perfectly match densities to earlier, but I think we should still aim for the same average vertex spacing.  

I think the 8k surface has sufficient resolution for faithfully representing any hippocampus we have seen (ignoring the dentate tail artifact from uniform coords), @jordandekraker would you agree? I would thus be in favour of dropping that, making the 8k the default, and adding a 512 vertices surface (ie resampling 12.5) to get ~2mm spacing. Or if you really think the 18k has value, could still include that, but I think it is overkill to keep that as the default.  

Here is an example den-512 surface (T1w data) for hipp and dentate:

<img width="1963" alt="Image" src="https://github.com/user-attachments/assets/6188d7f4-e1f0-4d34-b764-87292545d5bb" />

akhanf:
@jordandekraker forgot to bring up in the meeting whether we can just drop 18K as an output density-- thoughts? I think the 8k surface has sufficient resolution..

jordandekraker:
Yup, I agree we can make 8k the default. I'd like to keep the 18k around for things like ex-vivo or histology datasets, but no need to include it in the defaults. 

the 512 vertices looks good, I expect it'll be useful for things like fMRI but not so good for morphometry. Good to have it as an option!

akhanf:
Sounds good, I'll make a PR to change the default. There are a couple other minor things to address there too (missing coords metrics, and removing space wildcard from the spec file)..

PR #444, opened 2025-04-22
Atlas cli updates

akhanf:
This makes some cosmetic and validation updates for the atlas CLI. Biggest change now is the output densities available are based on the hippunfold-atlases repository, rather than hardcoded from the atlas creation resample choices. 

I've also added a den-512 atlas as per #443

### DIFF
--- a/hippunfold/plugins/atlas.py
+++ b/hippunfold/plugins/atlas.py
@@ -6,29 +6,29 @@ from pathlib import Path
 from typing import Any
 
 import attrs
+from appdirs import AppDirs
 from git import GitCommandError, Repo
 from snakebids import bidsapp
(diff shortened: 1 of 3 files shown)


akhanf:
note: although this points to an udpated hippunfold-atlases branch with the new den-512, the other densities shouldn't have changed (though we should confirm this..) - the files themselves are different likely due to numerical precision and encoding...

akhanf:
> note: although this points to an udpated hippunfold-atlases branch with the new den-512, the other densities shouldn't have changed (though we should confirm this..) - the files themselves are different likely due to numerical precision and encoding...

Looks like the atlas metric files do look a bit different between new and old -- I'm gonna run again end-to-end with this branch to make sure it's the latest..

PR #447, opened 2025-04-23
bugfix native surf affine

jordandekraker:
This fixes an issue where both left and right hippocampal surfaces were flipped and translated relative to corobl volumes. 

I believe that applying the affine to the grid wasn't working quite right, but applying it to the surface instead seems to do the trick in my tests.

### DIFF
--- a/hippunfold/workflow/scripts/gen_isosurface.py
+++ b/hippunfold/workflow/scripts/gen_isosurface.py
@@ -56,14 +56,12 @@ coords[sink_mask == 1] = 1.1
 src_sink_nan_mask = get_adjacent_voxels(sink_mask, src_mask)
 coords[src_sink_nan_mask == 1] = np.nan
 
-
 grid.cell_data["values"] = coords.flatten(order="F")
 grid = grid.cells_to_points("values")
-tfm_grid = grid.transform(affine, inplace=False)
 
 # Generate isosurface
 logger.info("Generating isosurface")
-surface = tfm_grid.contour(
+surface = grid.contour(
     [snakemake.params.threshold], method="contour", compute_scalars=True
 )
 logger.info(surface)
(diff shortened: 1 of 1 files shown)


akhanf:
hmm still curious why it doesn't affect my runs.. 
Let me give this a try this on my end before merging just in case..

jordandekraker:
for reference, I was getting this issue in the wetrun test-lowresMRI cases

akhanf:
my test is still running, but wondering if the issue is related to a pyvista new release.. makes me wonder @Dhananjhay if we should add version constraints to packages in env files so major releases don't break things.. Could the ^ version identifier we use in poetry also be used in conda env files?

Dhananjhay:
I've already version constrained the packages in the env files. The carrot (^) symbols used by poetry allow changes that do not modify the left-most non-zero digit. So ^1.23 translates to >=1.23.0, <2.0.0 and I have used the same logic in the conda env files. Looking at the pyvista env file:
```
name: pyvista
channels:
  - conda-forge
  - defaults
dependencies:
  - python=3.9
  - numpy>=1.21.2
  - nibabel>=3.2.1
  - pyvista>=0.44.2
  - vtk
  - scipy
  - pandas
```
We currently only put a lower limit. Should we add an upper limit as well?

akhanf:
> I've already version constrained the packages in the env files. The carrot (^) symbols used by poetry allow changes that do not modify the left-most non-zero digit. So ^1.23 translates to >=1.23.0, <2.0.0 and I have used the same logic in the conda env files. Looking at the pyvista env file:
> 
> ```
> name: pyvista
> channels:
>   - conda-forge
>   - defaults
> dependencies:
>   - python=3.9
>   - numpy>=1.21.2
>   - nibabel>=3.2.1
>   - pyvista>=0.44.2
>   - vtk
>   - scipy
>   - pandas
> ```
> 
> We currently only put a lower limit. Should we add an upper limit as well?

Yeah it's the upper limit I'm referring to -- this is what would prevent changes in new major releases from breaking our workflow.  We may then have to periodically update our environments to new versions (running wetrun tests to ensure validity when we do that) - looks like snakemake has also has a tool for updating versions of packages in conda envs: https://snakemake.readthedocs.io/en/stable/snakefiles/deployment.html#updating-environments

akhanf:
My test ran fine, so this is good to merge in I think. 
I'm pretty sure that a version change caused the issue, as that would explain why things would have been fine for me (as I've been using the same conda prefix for over a week)..

PR #449, opened 2025-04-23
Mackenzie/add modalities to quick

mackenziesnyder:
- add dsegtissue to hippunfold_quick 
- add image_modality dict with suffix (for dseg vs dsegtissue purposes and consistency) and use_derivatives
- if dsegtissue is the modality: add desc to file name, add dataset_description to the temp bids folder as per bids formatting error message, set derivatives flag to the temp bids folder
- dry run is successful

**Question:**
Is there a dseg test file I can use for wet run testing?

### DIFF
--- a/hippunfold/run_quick.py
+++ b/hippunfold/run_quick.py
@@ -73,10 +73,10 @@ def gen_parser():
         help="Execute a dry run without actually running the full pipeline.",
     )
     parser.add_argument(
-        "--hemi", 
-        required=False, 
-        choices= ["L","R"],
-        help="Brain hemisphere, tequired for dsegtissue modality. Choose between 'L' or 'R'."
+        "--hemi",
+        required=False,
+        choices=["L", "R"],
+        help="Brain hemisphere, tequired for dsegtissue modality. Choose between 'L' or 'R'.",
     )
 
     return parser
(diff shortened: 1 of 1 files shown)


akhanf:
if you want to do a wet-run test, you could use, e.g. sub-bbhist:
```
datalad install https://gin.g-node.org/hippunfold/multihist7
datalad get -d multihist7 multihist7/sub-bbhist/anat/*desc-tissue_dseg.nii.gz
```
Upon writing this I realize now we may actually need to add a CLI option for specifying the hemisphere, which would only be used when using modality=dsegtissue

mackenziesnyder:
> if you want to do a wet-run test, you could use, e.g. sub-bbhist:
> 
> ```
> datalad install https://gin.g-node.org/hippunfold/multihist7
> datalad get -d multihist7 multihist7/sub-bbhist/anat/*desc-tissue_dseg.nii.gz
> ```
> 
> Upon writing this I realize now we may actually need to add a CLI option for specifying the hemisphere, which would only be used when using modality=dsegtissue

Ok thanks! I will look into it

mackenziesnyder:
I ran a wet run test successfully by adding the `--skip-inject-template-labels` flag to the hippunfold-quick command. From the slack conversation I believe this change is to be left out of the actual commit based on the test data used - this pr should be good to test!

akhanf (approved):
Looks good to me!

PR #450, opened 2025-04-23
Add package pinnings

Dhananjhay:
Add upper and lower limit to packages inside the conda env to prevent breaking the pipeline.

### DIFF
--- a/hippunfold/workflow/envs/neurovis.yaml
+++ b/hippunfold/workflow/envs/neurovis.yaml
@@ -4,7 +4,7 @@ channels:
   - defaults
 dependencies:
   - python=3.9
-  - nilearn>=0.8.1
+  - nilearn>=0.8.1,<0.9.0
   - matplotlib=3.4.2
   - pandas>=1.2.0,<1.3.0
-  - seaborn>=0.11.2
+  - seaborn>=0.11.2,<0.12.0
--- a/hippunfold/workflow/envs/pyunfold.yaml
+++ b/hippunfold/workflow/envs/pyunfold.yaml
@@ -6,7 +6,7 @@ dependencies:
   - python=3.9
   - numpy=1.21.2
   - astropy>=4.0,<4.3.1
-  - nibabel>=3.2.1
-  - scikit-fmm>=2022.03.26
+  - nibabel>=3.2.1,<4.0.0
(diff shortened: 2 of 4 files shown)


Dhananjhay:
After pinning the dependencies in conda envs I was able to get past rule `postproc_boundary_vertices` but now it seems to be failing at rule `get_boundary_vertices`. I'm running the following command:
```
/hippunfold/run.py ../lowresMRI/ ../test-lowresMRI/ participant --participant-label 01 --use-conda --modality T1w --cores all
``` 

Error from the log file:
```
2025-04-24 12:55:50,684 - INFO - Loading surface from GIFTI...
Traceback (most recent call last):
  File "/local/scratch/test-lowresMRI/.snakemake/scripts/tmpmit2ffk1.get_boundary_vertices.py", line 15, in <module>
    surface, metadata = read_surface_from_gifti(snakemake.input.surf_gii)
  File "/local/scratch/hippunfold/hippunfold/workflow/lib/surface.py", line 26, in read_surface_from_gifti
    metadata = dict(vertices_darray.meta) if vertices_darray else {}
TypeError: 'GiftiMetaData' object is not iterable

```

akhanf:
ah I think the adding the pins for nibabel actually downgraded the version used quite a bit -- I am thinking let's take this opportunity to just upgrade packages, then pin.. I'm giving this a try with `snakedeploy update-conda-envs`, then `snakedeploy pin-conda-envs` 
(still running..)

Dhananjhay:
That's what I thought too, so I removed the upper limit from the `nibabel` package in the `pyvista.yaml` conda env and ran the pipeline to completion! Let's see what snakedeploy returns back but in the meantime I can pin `nibabel`'s version.

akhanf:
snakedeploy ends up upgrading everything to the latest versions, including latest python. Though it left some env files untouched (eg laynii, maybe because we only have the one khanlab build)?  

Anyways, for now since things are working, lets just merge this in, as upgrading everything would probably take some effort to make sure nothing breaks and do any minor repairs.. Can do that post-beta release..

Dhananjhay:
Sounds good, I can use snakedeploy post beta release and update the packages accordingly. This PR should be good to merge now, tried and tested with the lowresMRI dataset!

PR #455, opened 2025-04-25
AD trained model

jordandekraker:
work by @Bradley-Karat


hippunfold/workflow/rules/nnunet.smk
```diff
@@ -53,10 +53,19 @@ def get_model_tar():
 
     return (Path(download_dir) / "model" / Path(local_tar).name).absolute()
 
+def get_url():
+    if config["force_nnunet_model"]:
+        if config["force_nnunet_model"] == 'AD_T1w':
+            url='osf.io/ug3dz/download'
```
akhanf: this seems pretty hacky - why do we need to do it this way?

Bradley-Karat: I'm sure there is a better way to do it as it is a bit hacky of a work-around, but the download links for osf don't contain the model name like zenodo does (compare osf.io/ug3dz/download for the AD model vs the T1w model at zenodo.org/record/4508747/files/trained_model.3d_fullres.Task101_hcp1200_T1w.nnUNetTrainerV2.model_best.tar). In the nnunet.smk file we use the model name to parse the model which is defined in the config file for all other models as config["resource_urls"]["nnunet_model"][config["force_nnunet_model"]], but isn't for the AD model because of the OSF link not containing the .tar name. I just added in that fix so the config file will contain the model name and the get_url() function will contain the download url which doesn't have the AD model name. I think at the time we decided not to upload the model to Zenodo until it was finalized. Perhaps we could upload it now to zenodo to avoid this?

jordandekraker:
yeah its probably not great that we're parsing the model from the name of the URL - probably something we should work on in the future (or here). Either way, I just uploaded Brad's AD model to zenodo so it can be run just like the other models! just wetrun testing now to see how it looks

jordandekraker:
Also what's up with the test workflow not finding python 3.7? it doesn't look like its even in https://github.com/khanlab/hippunfold/blob/master/.github/workflows/python-testing.yml, so not sure why its being run...

akhanf:
yeah true at some point we can refactor code to avoid that.. but the easy thing to do right now is move to zenodo as you've done!

akhanf:
made a minor naming change for future proofing.. Also, note that by merging into master we won't have this in the dev-2.0.0 branch -- thinking maybe we just merge into dev-2.0.0 instead..

jordandekraker:
there are still a few issues with the .tar file - the data from fold_0 seems to be missing, and the folders have a different naming scheme than expected. @Bradley-Karat do you still have the original by any chance? also note that "tempmodel" shouldn't be in the tarred file, it just gets extracted to a folder named that. not sure if you renamed some of the other files/folders, but they should look something like this:

```
tempmodel/nnUNet/3d_fullres/Task301_ADNI_T1w_successful/Task301_ADNI_T1w_successful__nnUNetPlansv2.1/fold_0/nnUNetTrainer.model.pkl
tempmodel/nnUNet/3d_fullres/Task301_ADNI_T1w_successful/Task301_ADNI_T1w_successful__nnUNetPlansv2.1/fold_0/nnUNetTrainer.model
```

Bradley-Karat:
ah good catch, not sure why fold_0 was missing from that one - I must've uploaded the wrong one. The model with fold_0 can be downloaded here: https://osf.io/kch9v/download

jordandekraker:
OK the new model should be available, could someone please run a quick test to ensure its publicly accessible and not just to me?

Can alse edit in any additional information on the zenodo page, i tried to make it similar to the neonatal project: https://zenodo.org/uploads/15297857

akhanf:
Do we want to merge this into dev-2.0.0 instead of master? Or we could cherry-pick from dev-2.0.0 and keep this merging into master as is..

jordandekraker:
i was thinking the latter since there's nothing specific to dev here, and technically the AD model has already been available for a while via this branch & docker build prior to the PR. But i'm fine either way

akhanf (approved):

akhanf:
Yes I agree, merging it in!

PR #456, opened 2025-04-27
Dockerize conda

Dhananjhay:
Replace relative path to the run.py executable with its absolute path, and remove default flags, i.e., `--use-conda` and `--conda-prefix /src/conda-envs`. The latter is a quick hack and the user would have to pass those default flags from the command line but in future it would be more robust to integrate these flags in the pipeline from inside the docker image.


akhanf:
actually I think I have a fix for this now, using a default profile, will make a separate PR to test out..

akhanf:
ok I was able to get things ironed out in #457, so we can close this without merging

PR #457, opened 2025-04-27
fixes for docker build

akhanf:
- simplify entrypoint with conda activate instead of conda run
- typo conda-env to conda-envs in one of the env creation run
- uses a profile for setting default conda args
- also changes dockerfile from dos to unix (cosmetic)


akhanf:
Running an end to end test (so far looks fine) then will merge this in

PR #458, opened 2025-04-28
Setup CI for Conda recipe

Dhananjhay:
This PR adds a new workflow which builds and pushes the latest version of hippunfold on custom khanlab anaconda repo.

### DIFF
--- a/.github/workflows/push_conda.yml
+++ b/.github/workflows/push_conda.yml
@@ -19,15 +19,17 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
-      - name: Set up Conda
+      - name: Setup Mambaforge
         uses: conda-incubator/setup-miniconda@v2
         with:
-          miniconda-version: "latest"
+          miniforge-variant: Mambaforge
+          miniforge-version: latest
+          use-mamba: true
           auto-activate-base: true
 
       - name: Install dependencies
         run: |
-          conda install -y anaconda-client
+          mamba install -y conda-build anaconda-client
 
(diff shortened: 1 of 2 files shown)


Dhananjhay:
- Caches Conda build artifacts (conda-bld) to speed up repeated builds.
- Uses Miniforge3 and the libmamba solver for faster and more reliable dependency resolution.
- Dynamically updates the ref and recalculates the sha256 checksum in meta.yaml to match the latest commit on the dev-v2.0.0 branch.
- Builds the Conda package using conda mambabuild.
- Uploads the package to the Khanlab Anaconda channel, replacing any previous versions.

akhanf:
I see this is hard-coded for 2.0.0 -- would we need to also update the version bumping actions accordingly?

Dhananjhay:
Good point, I'll add the logic to bump the version number!

Dhananjhay:
> I see this is hard-coded for 2.0.0 -- would we need to also update the version bumping actions accordingly?

So right now the way this action is setup to push a new conda package for hippunfold version 2.0.0 every time there's a new push to the dev-v2.0.0 branch. The main idea behind this was to support new patches and ongoing development in the dev-branch. To make more generalizable, we'd have to change the logic in both the push_conda.yaml and bump_version.yaml actions.

akhanf:
ah yes, ok lets keep this current behaviour for now while dev-2.0.0 is still under active development, and we can make another PR later as needed when we release..

Dhananjhay:
Agreed! We can make another PR that's more general later after we release v2.0.0!

Issue #461, opened 2025-05-02
Issues using wget on non-linux OS


Ian-Goodall-Halliwell:
In testing the new version (2.0), using the conda installation instructions on the wiki, the instructions use the command 

`wget https://www.dropbox.com/s/mdbmpmmq6fi8sk0/hippunfold_test_data.tar`

This causes an error on macOS, and likely will on windows as well, due to wget not being a preinstalled package on those systems by default.

By my testing, 

`curl -L https://www.dropbox.com/s/mdbmpmmq6fi8sk0/hippunfold_test_data.tar -o hippunfold_test_data.tar`

works, and should be supported on most OS.

After running this, and continuing with the installation steps, the first run of the pipeline ran into many issues because wget was being used to download many of the model files.

akhanf:
Quick fix for mac could be to ensure the conda wget package is a dependency in hippunfold, but we should probably migrate to using curl instead (also need to ensure the curl conda package is in the hippunfold top-level environment)..

PR #465, opened 2025-05-07
--use-conda as default through default profile

akhanf:
This sets --use-conda as the default. However, it breaks our github actions for dry run testing, since those do not install from conda.. 

We should probably update our actions to use conda in any case - @Dhananjhay are you able to look at that?


Dhananjhay:
Sure!

akhanf:
actually, looks like the error is specific to the old snakemake (version ~7) that gets used with python <3.11, which has mamba set-up as the default front-end.. the tests pass with python 3.11 or greater.. 

I am thinking maybe it is best to restrict to a modern snakemake version (also effectively restricts to python >= 3.11), not only for this issue, but also so we don't have to support the historical cluster execution api in snakemake (snakemake <8)

akhanf:
just saw your commit now - good idea to do this in any case, thanks!

Dhananjhay:
> actually, looks like the error is specific to the old snakemake (version ~7) that gets used with python <3.11, which has mamba set-up as the default front-end.. the tests pass with python 3.11 or greater..
> 
> I am thinking maybe it is best to restrict to a modern snakemake version (also effectively restricts to python >= 3.11), not only for this issue, but also so we don't have to support the historical cluster execution api in snakemake (snakemake <8)

Sounds good! Updating pyproject.toml should be straightforward. For Conda, we don't need to change anything because by default the newest stable build, Python v3.11, is installed in a newly created env. I was also thinking of switching to mamba for installation and addressing issue #461. Probably best to open another PR for that though.

jordandekraker:
i think this is good to merge, yes?

Dhananjhay:
I just tried out the same logic for the `autoafids` package and it works fine when using docker but not with the corresponding conda package, so maybe hold off merging this PR until I've ran a test with these changes using the `hippunfold`'s conda package.

Dhananjhay:
Yea, it doesn't work when working with its corresponding conda package. Let me see what's the bottleneck here.

Dhananjhay:
Solved the issue with the conda package, it's good to merge now!

PR #466, opened 2025-05-07
wget replaced by curl -L and -O

jordandekraker:
simple solution for https://github.com/khanlab/hippunfold/issues/461


akhanf:
The rule doesn't use a conda environment, should perhaps make one with curl (there is a conda package for it i believe) as the only dependency so it works regardless of user home environment (or OS).

Dhananjhay:
I've added a new conda env with only curl dependency and injected it in the rules wherever necessary. Also, slightly changed the curl command to use `-o` instead of `-O`.  
`-O` → use original filename from server
`-o` <name> → set your own output filename

Issue #469, opened 2025-05-13
unzip: command not found in Docker container on Windows


Ian-Goodall-Halliwell:
I am following the guide on the wiki to run hippunfold 2.0 on windows, but when running:

`docker run -it --rm -v c:\Users\ian\Downloads\ds002168:/bids -v c:\Users\ian\Downloads\ds002168_hippunfold:/output khanlab/hippunfold:dev-v2.0.0 /bids /output participant --modality T1w -p --cores 4`

I receive the following error:

`/usr/bin/bash: line 1: unzip: command not found
RuleException:
CalledProcessError in file "/src/hippunfold/workflow/rules/download.smk", line 20:
Command 'source /opt/conda/bin/activate '/src/conda-envs/ea11a0998b7bd1b02424c869b9113f8e_'; set -euo pipefail;  curl -L 'https://files.ca-1.osf.io/v1/resources/v8acf/providers/osfstorage/65395c1613d27b122a94ca09/?zip=' -o temp.zip &&  unzip -d /root/.cache/hippunfold/template/upenn temp.zip' returned non-zero exit status 127.
Exiting because a job execution failed. Look below for error messages
WorkflowError:
At least one job did not complete successfully.
Traceback (most recent call last):

  File "/opt/conda/envs/snakebids-env/lib/python3.12/site-packages/snakemake/executors/local.py", line 232, in spawn_job    subprocess.check_call(cmd, shell=True)

  File "/opt/conda/envs/snakebids-env/lib/python3.12/subprocess.py", line 413, in check_call
    raise CalledProcessError(retcode, cmd)

[...]

Issue #384, opened 2025-03-10
run_inference fails on CBS basic for the --use-conda flag


mackenziesnyder:
- When the --use-singularity flag is specified on cbs basic run_inference works
- command run: `hippunfold ~/Desktop/lowresMRI/ /localscratch/hipp_output participant --use-conda --modality T1w --cores all`

Error message: 

```
/usr/bin/bash: line 1: 32702 Killed                  nnUNet_predict -i tempimg -o templbl -t Task101_hcp1200_T1w -chk model_best -tr nnUNetTrainerV2 --disable_tta &> logs/sub-01/sub-01_hemi-L_space-corobl_nnunet.txt
[Mon Mar 10 10:13:16 2025]
Error in rule run_inference:
    jobid: 0
    input: work/sub-01/anat/sub-01_hemi-L_space-corobl_desc-preproc_T1w.nii.gz, /localscratch/.cache/hippunfold/model/trained_model.3d_fullres.Task101_hcp1200_T1w.nnUNetTrainerV2.model_best.tar
    output: work/sub-01/anat/sub-01_hemi-L_space-corobl_desc-nnunet_dseg.nii.gz
    log: logs/sub-01/sub-01_hemi-L_space-corobl_nnunet.txt (check log file(s) for error details)
    conda-env: /local/scratch/hipp_output/.snakemake/conda/3bf59d51ad8aa8841da91b382341fb82_
    shell:
        mkdir -p tempmodel tempimg templbl && cp work/sub-01/anat/sub-01_hemi-L_space-corobl_desc-preproc_T1w.nii.gz tempimg/temp_0000.nii.gz && tar -xf /localscratch/.cache/hippunfold/model/trained_model.3d_fullres.Task101_hcp1200_T1w.nnUNetTrainerV2.model_best.tar -C tempmodel && export RESULTS_FOLDER=tempmodel && export nnUNet_n_proc_DA=4 && nnUNet_predict -i tempimg -o templbl -t Task101_hcp1200_T1w -chk model_best -tr nnUNetTrainerV2 --disable_tta &> logs/sub-01/sub-01_hemi-L_space-corobl_nnunet.txt && cp templbl/temp.nii.gz work/sub-01/anat/sub-01_hemi-L_space-corobl_desc-nnunet_dseg.nii.gz
        (one of the commands exited with non-zero exit code; note that snakemake uses bash strict mode!)

Exiting because a job execution failed. Look above for error message
WorkflowError:
[...]

PR #472, opened 2025-05-16
Refactor unzip command tool 

Dhananjhay:
This PR fixes issue #469 by refactoring the `download_extract_template` rule to use Python’s standard `zipfile` library instead of the `unzip` command-line tool.

### DIFF
--- a/hippunfold/workflow/rules/download.smk
+++ b/hippunfold/workflow/rules/download.smk
@@ -25,7 +25,7 @@ rule download_extract_template:
 
         # 1) download the zip
         zip_path = os.path.join(outdir, "temp.zip")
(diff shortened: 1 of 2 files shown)


akhanf (changes requested):
The curl conda env is no longer needed, nor is it used since the run directive doesn't use the env and just uses the workflow env. That said, this is a fairly long run block, so for the sake of readability it would be preferred to put it in a script.

Dhananjhay:
I've migrated the logic to a python script. I haven't tested these changes out since I'm working on issue #384 but I don't suppose anything would be broken. Would prompt when local testing gives a green signal!

akhanf:
just tested this out, works fine for me so I will merge this in..

Issue #464, opened 2025-05-06
moths in the DG


jordandekraker:
As raised by Matt Glasser, there seems to be occasional issues with holes in the dg:

![Image](https://github.com/user-attachments/assets/62c0281c-8bae-4bf3-b13c-7379d3ceb34a)

In our high-res and low-res cases, and in a few HCP subjects, I haven't been able to replicate this issue, but I will try with the exact same data (HCP-100307) once I track down a copy. 

Note: Ali suggested CLI option to try:

--laminar-coords-res-dentate 

the default is 0.1mm, but if you bring this down (0.05) 

This is what I would try as well, but it is sort of a brute force solution and shouldn't really be needed since the DG is typically ~0.3mm thick

glasserm:
It didn't help, unfortunately.

akhanf:
Took a look at this today and found the underlying issue --

After our template shape injection, we have always been adding back in the Cyst label from the nnunet segmentation (since the Cyst isn't labelled with the registration-based shape injection)..  However, for the dentate coords, this Cyst label seems to be clobbering the dentate GM label at times, especially since for the dentate we inject at an upsampled (0.1mm) resolution, so the (0.3mm) Cyst label is much coarser. Before dev-2.0, we weren't solving coords for dentate and simply injecting the template coords, which is why this wasn't happening before.. 


Here is an example showing 100307, with the `desc-postproc` (ie labels after injection and adding in Cyst), where Cyst label is yellow, and dentate surface outlined in blue.

<img width="608" alt="Image" src="https://github.com/user-attachments/assets/cbd5bab6-b4d8-43c3-aa2e-7399c28ffd1f" />

and surface only:

<img width="608" alt="Image" src="https://github.com/user-attachments/assets/d38a2748-df29-4e34-afbc-ad23c7e0af56" />


Initially I was thinking of just skipping this step for dentate, but I think the better solution is to only allow the Cyst label to clobber the SRLM. This is where we anatomically expect it anyhow, and ensures we don't end up breaking any topology. 
 In the first image, it is clobbering the dentate and making a hole, and the fix is below.


<img width="608" alt="Image" src="https://github.com/user-attachments/assets/bfb37385-23cb-4be7-aad7-5968c1943c12" />

[...]

PR #476, opened 2025-05-26
Fix dentate holes

akhanf:
Deals with large dentate holes (#464 ) caused by nnunet Cyst label.

This fix ensures re-inserted cyst label only overwrites SRLM, which should ensure it doesn't alter the topology present in the injected segmentation.

Note: There still seem to be some tiny holes that are present (e.g. in hcp subject 100307), but these seem to be related to the quality of the dentate src/sink labels defined in the template..


akhanf:
here is a before and after for subject 100307:

image


image

PR #478, opened 2025-05-28
removed unused autotop container & unused resources

jordandekraker:
simple find&replace all instances of 
```
    container:
        config["singularity"]["autotop"]
```

Also removed some resources of old surface files


akhanf:
May as well also remove the conda_env wrapper if we are removing all the container directives.

akhanf:
Also could update the docs to remove any use of --use-singularity or --use-apptainer since those wouldn't be applicable anymore

Dhananjhay:
I've removed the `conda_env` wrapper logic and some remaining `container:` directives from the codebase. Re: removing `--use-singularity` from the documentation, seems like it's mentioned only in `contributing.md`. Should I update it in such a way that we ask users to run hippunfold within the context of conda and not poetry + singularity?

akhanf:
> I've removed the `conda_env` wrapper logic and some remaining `container:` directives from the codebase. Re: removing `--use-singularity` from the documentation, seems like it's mentioned only in `contributing.md`. Should I update it in such a way that we ask users to run hippunfold within the context of conda and not poetry + singularity?

Just having a look now -- yeah that contributing page would be almost entirely changed.. could update docs in another PR, maybe makes sense to hold off until the uv PR is merged (or pixi if we try that too)..

PR #481, opened 2025-06-01
Add adni model

Dhananjhay:
In reference to PR (AD trained model #455). Hippunfold v2.0.0 was lacking the ADNI model.


akhanf (approved):
looks good to me

PR #482, opened 2025-06-02
ieeg electrode annotation and surface-labelling

akhanf:
This was a request from some of our clinical collaborators here (@arunthurai is helping facilitate this), for being able to identify what hippocampus label is associated with each electrode contact, and provide some means for visualizing these in group space. It allows the user to define another image as the anatomical reference space too.. 

@jordandekraker I know you have a hippomaps example using ieeg data, but the use-case and input data are a bit different. In your example you're calculating something per electrode using non-bids data, mapping it to volume-space (with a radius around each electrode), and then mapping it to the hippocampus surfaces.  Here we are making use of a BIDS electrodes tsv file as input, and simply finding the closest hippocampal vertex, and annotating the tsv file, along with creating gifti outputs to indicate the closest vertex for each electrode contact. 

Anyways happy to discuss this further in case we want to expand on this (leaving it as a draft for now).. 

Happy to also discuss general issues of scope of hippunfold -- I see this PR as within scope of hippunfold (and not part of a separate tool) because it is relying on BIDS spec inputs (e.g the ieeg electrodes tsv file), and producing outputs that are congruent with the existing hippunfold outputs. I.e. something that provides some minimal mapping which users could then make use of in their more involved analyses.. 

@arunthurai if you want to try it out, something like this should work (needs the --path options because of the peculiarities of your bids dataset):

```
hippunfold hippunfold_seeg/bids  out_folder participant -p  --modality T1w  --filter-T1w acquisition=MP2RAGEpostproc  --path-electrodes hippunfold_seeg/derivatives/sub-{subject}/sub-{subject}_space-T1w_SEEGA.tsv   --path-electrodes_ref hippunfold_seeg/bids/sub-{subject}/ses-pre/anat/sub-{subject}_ses-pre_run-02_T1w.nii.gz  --use-electrodes
```


arunthurai:
thanks ali, ran successfully for the patients in that cohort. not sure if it's relevant but for the patients without 7T (i.e. the clinical 1.5T and the two with a 3T scan from robarts) i had to set the output-density for the vertices to be at native resolution

akhanf:
@Derbedog @jclauneuro, could be relevant to your work -- remind me to show you this feature at some point, would be good to get your feedback and merge it in..

PR #483, opened 2025-06-04
avoid the remove-rule by using temp() for all

akhanf:
this can work now since a bugfix in snakemake v9.3.0, which retains files that are included as targets


akhanf:
tested this out locally so this should be good to merge, simplifies the workflow by removing this hacky deletion rule

jordandekraker:
beautiful! I haven't tested locally yet but just wanted to make sure: does this work if we include `--output_density native`? (i.e. is `temp()` smart enough to hadle this?)

akhanf:
Yep, I tested it out with --output-density native

I've confirmed that temp() is now overriden if the file is also included as a target file.. 

Note, if you run the pipeline with one set of densities then re-run with another, it generally won't clean-up the old ones if they aren't in the new dag at all, but I think that's fine..

PR #484, opened 2025-06-05
Pixi package management

akhanf:
Trying out pixi here.. 
Their build process isn't quite developed yet, but I am thinking it still might be preferred over other options (poetry, uv), since it can handle conda dependencies. 

They do also have a way to pack-up and distribute as a tar file or as an installer -- this could be a nice way of distributing hippunfold or snakebids apps in general.. https://pixi.sh/latest/deployment/pixi_pack/

Also, I couldn't get a console script defined so I've made use of pixi tasks instead..


jordandekraker:
This worked like a charm for me! So fast, and the docs look solid. 

My only question is whather we should remove the conda install instructions & just use this, or keep both. @akhanf any thoughts on that?

akhanf:
this looks complete so merging it in too

akhanf:
> This worked like a charm for me! So fast, and the docs look solid.
> 
> My only question is whather we should remove the conda install instructions & just use this, or keep both. @akhanf any thoughts on that?

I think fine to keep conda instructions too, as more people are familiar with conda

PR #486, opened 2025-06-09
updates from master to dev

jordandekraker:
This is so we can sort out conflicts ahead of merging the dev branch


akhanf:
Ah are you sure you're not breaking things with this? We have been cherry picking from master to avoid issues (eg the ADNi model PR)..

I don't think merging dev2.0 into master is ultimately the best way forward for release,  rather can just make it the main branch.. there will be tons of things to merge and doesn't really make sense to do it (will only create messy history with no real gain)

jordandekraker:
no i'm not sure that this is the optimal way, but i figured it wouldn't be a big deal since the number of commits in master is relatively small, and the "Files changed" listed here is also pretty small/simple (probably because most of the changes were already cherry-picked). For a cleaner history I could have perhaps used fast-forward...

akhanf:
It's not the history in this PR i'm worried about, it's what happens when we try to merge dev2.0.0 back into master.. since dev is essentially acting as the new master I think it makes sense to just keep it as is for the new release.. we could rename `dev-2.0.0` to `main` as well (since it's probably a good time to also update naming to main instead of master).. can keep the old master branch around in case we ever need it, perhaps renaming it to avoid confusion (master-1.x?)..

jordandekraker:
hmm yeah that would be fine but I was kinda hoping to get some stats, eg. number of commits and contributors, which only shows up on the default branch. This will show up for all time if we merge into master, but i'm not sure what will happen if we switch dev-v2.0.0 to the default branch (and rename to main)

akhanf:
yep that all happens for whatever the default branch is, so would track the new main when we make that change..

PR #503, opened 2025-07-18
updated approach for open isosurface creation

jordandekraker:
In principle this is a pretty simple fix, we just take out the extra dilated nan mask and then find a way to remove the unwanted components of the surface. This would be better than postprocessing to patch holes since it allows for a smooth&curvy surface. 
image
Here i tried keeping the second largest connected component but it must be connected somewhere because that effectively removes the whole thing
Would also need to remove from dg:
image
some other masking of the bg might work, will look into it later

### DIFF
--- a/hippunfold/config/snakebids.yml
+++ b/hippunfold/config/snakebids.yml
@@ -542,6 +542,7 @@ laplace_labels:
     IO:
       gm:
         - 1
+        - 8
       src:
         - 2
         - 4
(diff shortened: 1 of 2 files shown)


jordandekraker:
I think i was tired when i tried the "second largest CC" mask last week. An easier way is to get the largest CC of invalid vertices (i.e. those outside of the coords mask), and discard those. This keeps the surface intact in holes (which are islands of invalid vertices).

I think this is pretty simple and elegant, doesn't introduce any new parameters or dependencies, and adds negligible processing time. 

Let me know how it works in your test cases @akhanf

akhanf:
Sounds good will give it a whirl. Logic seems sound to me.

akhanf:
merging this in -- fyi I changed the title to describe the change

Issue #498, opened 2025-07-04
laplace-beltrami occasionally failed


jordandekraker:
one of my 49 subjects failed for one hemisphere, but i can't work out why (yet). The earliest issue seems to be that Laplace-Beltrami for both AP and PD are completely wrong, even though the boundary conditions look good. I did note that there are some small holes in the surface, but i don't think that would produce this result.

AP coords:

<img width="1426" height="935" alt="Image" src="https://github.com/user-attachments/assets/83668d50-f595-4a63-824d-3425e26d3ec4" />

AP boundary conditions:

<img width="1401" height="961" alt="Image" src="https://github.com/user-attachments/assets/d23eaacb-cf1f-43c8-a8a8-001f136558b6" />

Note that the PD coords also seem to be similarly stratified along the sagittal direction:

<img width="1411" height="928" alt="Image" src="https://github.com/user-attachments/assets/51d2af4e-4cf2-4a73-b78d-4cc90a2cf704" />

akhanf:
I've also seen this now in a handful of cases , e.g. HCP subject 129028, happens for left hipp.

Looking at the laplace beltrami logs, seems like it is having a hard time solving,  the rank warning doesn't come up in the successful cases.. 

I also couldn't find anything noticeably wrong in the surface mesh and src/sink inputs, so might need to dig in and see what the matrix it is trying to solve looks like.. 

```
/tmp/hippunfold/.snakemake/scripts/tmpfvpi0v5r.laplace_beltrami.py:102: MatrixRankWarning: Matrix is exactly singular
  solution[free_indices] = sp.linalg.spsolve(free_laplacian, free_b)
```

akhanf:
ok looking at this case, 
1) the matrix being singular is caught as an exception by the script, setting everything to zeros, otherwise we have all nans (neither approach makes the data usable btw)..
2) the underlying issue in this case (and my bet for all these cases) are holes in the surface, though it's not clear to me why this causes a singularity?? 

Note: in wb_view you want to turn on backface culling to see these holes.

<img width="646" height="438" alt="Image" src="https://github.com/user-attachments/assets/4968326f-c1d7-421a-8a28-1a9b0d80006a" />

jordandekraker:
yup, i'd noticed these kinds of holes as well. obviously fixing the source is preferrable to modifying the L-B solver since that will also fix downstream issues with things like thickness calculation. 

[...]

PR #504, opened 2025-07-20
pymeshfix

jordandekraker:
using pymeshfix

https://pymeshfix.pyvista.org/examples/repair_planar.html


akhanf:
added to the #498 issue, but this was causing issues in dentate (for the one subject I tried), so trying out all subjects without.. 

Note: I've also included an unrelated fix for ants unfoldreg to clamp surface metrics, this was needed to get the test subject HCP 129018 to run to completion..

akhanf:
This isn't being used anymore (achieved robustness through other means) so closing without merging..

PR #506, opened 2025-07-23
fix CLI arg parsing for laminar_coords_res

akhanf:
the destination variable wasn't being populated so the argument had no effect


akhanf:
merging this simple bugfix in

PR #507, opened 2025-07-23
enforces connected-components in src/sink labels

akhanf:
by keeping the largest connected component, and relabelling smaller components based on their neighbours


jordandekraker (commented):
i see you added networx to the pyvista env, but it doesn't look like its actually being used?

akhanf:
ah the import is just down inside the function

PR #509, opened 2025-07-24
Hole fill radius update, hole reporting in get_boundary_vertices

akhanf:
-updates hole fill radius (fixes the remaining surface issue in HCPUR100)
-adds some reporting on holes in the log for get_boundary_vertices, including the holes as a distinct color in the resulting label gifti too (this label is only for qualitative purposes, doesn't get used otherwise..)


akhanf:
Merging this in, as it doesn't cause any issues -- I'm still seeing some LB solver issues, but actually it looks like just using some light regularization for the solver does the trick, will do that in a separate PR..

PR #510, opened 2025-07-27
adds a small amount of tikhonov regularization

akhanf:
should fix any remaining solver issues (cause by holes or other unnoticeable mesh defects)


akhanf:
Confirmed this works fine on hcp ur100, no adverse side effects and now finally have all running to completion.

PR #514, opened 2025-09-04
Check for atlas-sync

Dhananjhay:
Instead of opening a socket connection to github.com:443, we just use `git ls-remote url` for quickly inspecting the contents of a remote repository and/or verifying its accessibility. Socket logic seemed like an overkill; it was failing on server where git clone works.


akhanf (approved):
thanks looks good to me!

PR #515, opened 2025-09-08
deps + docs for cluster execution

jordandekraker:
This PR adds `snakemake-executor-plugin-cluster-generic` as a dependency which seems to be needed for cluster execution. It also provides a new "HCP" section in the Docs to explain how to use this feature. 

TODO: 
- [ ] test these instructions at the BIC (SGE)
- [ ] test these instructions on computecanada (SLRUM)


docs/usage/HPC.md
```diff
@@ -0,0 +1,118 @@
...
+
+---
+
+## Before you start
+
+When running on a cluster, please check the following:
+
+* **Installation path**: Make sure HippUnfold is installed in a directory that is accessible from all compute nodes (e.g., shared network storage).
+
+* **Containers**: If you are using Apptainer/Singularity or Docker, confirm that the runtime is available on compute nodes.
```
akhanf, line 17: can you actually use hippunfold called from a container along with a executor plugin? I didn't know there was a way for the remote jobs to call the container properly.. Unless you've tested this we could remove this statement (or say you cannot use the hippunfold container with executor plugins)

docs/usage/HPC.md
```diff
@@ -0,0 +1,118 @@
...
+
+* **Conda environments**: If using conda, ensure your environment directory is on a network-accessible path.
```
akhanf, line 19: could remind them that by default it will be in the output hippunfold folder (in .snakemake), but can be overriden with `--conda-prefix`

docs/usage/HPC.md
```diff
@@ -0,0 +1,118 @@
...
+jobs: 20
+cores: 16
+max-threads: 16
+latency-wait: 120
+conda-frontend: mamba
+conda-prefix: /data/mica3/.snakemake/conda
+precommand: "source /host/cassio/export03/data/opt/miniconda3/etc/profile.d/conda.sh"
+default-resources:
+  - mem_mb=8000
+  - time_h=6
```
akhanf, line 57: is time_h setting a default resource for each rule? seems really high if so.. but if it's the grouped rule then more reasonable..

docs/usage/HPC.md
```diff
@@ -0,0 +1,118 @@
...
+with optional flags like `-q all.q` or `-q mica.q` (used in the [MICA Lab](https://mica-mni.github.io/)).
+
+When running HippUnfold on SGE, replace the `--cores` option with `--profile` to point Snakemake to an SGE profile. For example, a profile might be stored at:
+
+`~/.config/snakemake/sge/config.yml`
+
+```yaml
+executor: cluster-generic
+cluster-generic-submit-cmd: "qsub -cwd -q mica.q"
+jobs: 20
```
akhanf, line 48: this would limit you to 20 jobs at a time, is that the intention?

docs/usage/HPC.md
```diff
@@ -0,0 +1,118 @@
...
+
+SLURM is another widely used cluster manager (e.g., on all **Compute Canada** systems). Jobs are usually submitted with:
+
+```bash
+sbatch <your command>
+```
+
+As with SGE, HippUnfold jobs can be distributed using a Snakemake profile for SLURM. A minimal example:
+
+`~/.config/snakemake/slurm/config.yml`
```
akhanf, line 80: using a profile for the slurm executor is a bit overkill in my opinion -- really the only things you need to set are the executor, cores and jobs, right? clsuter-cancel and clsuter-status should already be managed by the plugin ..

docs/usage/HPC.md
```diff
@@ -0,0 +1,118 @@
...
+Submit HippUnfold with:
+
+```bash
+hippunfold <your usual arguments> --profile slurm
+```
+
+You may also need to customize `sbatch` options (e.g., partitions, QoS, GPUs). To do this, define a submission command in your profile, for example:
+
+```yaml
+cluster-generic-submit-cmd: "sbatch --partition=compute --time={resources.time_h}:00:00 --mem={resources.mem_mb}M"
```
akhanf, line 106: you're combinign sbatch with cluster generic here, but there are slurm executor options to change this, we should really just refer the user to: https://snakemake.github.io/snakemake-plugin-catalog/plugins/executor/slurm.html for the detailed options for slurm

pyproject.toml
akhanf, line 1: can also add the slurm plugin

jordandekraker:
I was never able to get this running on our SGE cluster. After a few days of fiddling with parameters and cache/config dirs I gave up and am simply doing `qsub singulairty run hippunfold.sif <IN> <OUT> participant` in batches. 

If I wasn't able to get this running then I don't really think we can support it as a feature, so maybe we should simply close this PR

PR #516, opened 2025-09-08
WIP simplifying Docs

jordandekraker:


docs/getting_started/conda_advanced.md
akhanf, line 1: This section needs some more significant editing (mostly removing things like Poetry and anything already stated elsewhere) - I also think it is better to instead call it something like "Setting Up a Dev Environment", and moving it down to another -- having it called Conda(Advanced) makes it confusing between the other Conda section..  

I can give this a crack

akhanf:
this looks good to me, aside from my comment on the advnaced section, made just a few minor typo corrections

jordandekraker:
I'm happy to merge this for now, but if you guys are really happy with pixi then maybe you could add it to the Docs here instead of poetry? That would help me get started using pixi as well

I'm also removing the HPC page since that is in https://github.com/khanlab/hippunfold/pull/515 and may get deleted

PR #517, opened 2025-09-23
untar of model in separate rule; only done once

jordandekraker:
I've been finding the run_inference step still pretty slow, especially on clusters where read/write is the major bottleneck. The previous behaviour of this step would untar the nnunet model into a local temporary directory, and then load it into memory. Here we will untar it in the cache directory. This will only need to be done once instead of on a per-subject basis. I hope that this will speed things up a bit. 

If the untar is interrupted, we should be able to --rerun-incomplete, but there could be some concern if it is interrupted so the model dir exists and is called by a different hippunfold run though.

I'm also replacing `cp` with `ln -s`. I think this should be fine since we're not supporting Windows outside of docker anyway

### DIFF
--- a/hippunfold/workflow/rules/nnunet.smk
+++ b/hippunfold/workflow/rules/nnunet.smk
@@ -106,12 +106,12 @@ def get_cmd_copy_inputs(wildcards, input):
     in_img = input.in_img
     if isinstance(in_img, str):
         # we have one input image
-        return f"ln -s {in_img} tempimg/temp_0000.nii.gz"
+        return f"cp {in_img} tempimg/temp_0000.nii.gz"
     else:
         cmd = []
         # we have multiple input images
         for i, img in enumerate(input.in_img):
-            cmd.append(f"ln -s {img} tempimg/temp_{i:04d}.nii.gz")
+            cmd.append(f"cp {img} tempimg/temp_{i:04d}.nii.gz")
         return " && ".join(cmd)
 
 
--- a/hippunfold/workflow/rules/nnunet.smk
+++ b/hippunfold/workflow/rules/nnunet.smk
@@ -111,16 +111,25 @@ def get_cmd_copy_inputs(wildcards, input):
         return " && ".join(cmd)
 
 
+rule unpack_nnunet_model:
+    """ Unpack nnunet model tar to temp folder to check contents"""
+    input:
+        get_model_tar(),
+    output:
+        get_model_tar() + "unpacked",
+    shell:
+        "mkdir -p {output} && tar -xf {input} -C {output}"
+
+
 rule run_inference:
     """ This rule uses either GPU or CPU .
     It also runs in an isolated folder (shadow), with symlinks to inputs in that folder, copying over outputs once complete, so temp files are not retained"""
     input:
         in_img=get_nnunet_input,
-        model_tar=get_model_tar(),
+        model_dir=get_model_tar() + "unpacked",
     params:
         cmd_copy_inputs=get_cmd_copy_inputs,
         temp_lbl="templbl/temp.nii.gz",
-        model_dir="tempmodel",
         in_folder="tempimg",
         out_folder="templbl",
         task=parse_task_from_tar,
@@ -159,15 +168,13 @@ rule run_inference:
     shell:
         #create temp folders
         #cp input image to temp folder
-        #extract model
         #set nnunet env var to point to model
         #set threads
         # run inference
         #copy from temp output folder to final output
-        "mkdir -p {params.model_dir} {params.in_folder} {params.out_folder} && "
+        "mkdir -p {params.in_folder} {params.out_folder} && "
         "{params.cmd_copy_inputs} && "
-        "tar -xf {input.model_tar} -C {params.model_dir} && "
-        "export RESULTS_FOLDER={params.model_dir} && "
+        "export RESULTS_FOLDER={input.model_dir} && "
         "export nnUNet_n_proc_DA={threads} && "
         "nnUNet_predict -i {params.in_folder} -o {params.out_folder} -t {params.task} -chk {params.chkpnt} -tr {params.trainer} {params.tta} &> {log} && "
         "cp {params.temp_lbl} {output.nnunet_seg}"
--- a/hippunfold/workflow/rules/nnunet.smk
+++ b/hippunfold/workflow/rules/nnunet.smk
@@ -116,7 +116,11 @@ rule unpack_nnunet_model:
     input:
         get_model_tar(),
     output:
-        get_model_tar() + "unpacked",
+        Path(download_dir)
+        / "model"
+        / config["resource_urls"]["nnunet_model"][config["force_nnunet_model"]]
+        if config["force_nnunet_model"]
+        else config["resource_urls"]["nnunet_model"][config["modality"]] + "unpacked",
     shell:
         "mkdir -p {output} && tar -xf {input} -C {output}"
 
@@ -126,7 +130,15 @@ rule run_inference:
     It also runs in an isolated folder (shadow), with symlinks to inputs in that folder, copying over outputs once complete, so temp files are not retained"""
     input:
         in_img=get_nnunet_input,
-        model_dir=get_model_tar() + "unpacked",
+        model_tar=get_model_tar(),
+        model_dir=(
+            Path(download_dir)
+            / "model"
+            / config["resource_urls"]["nnunet_model"][config["force_nnunet_model"]]
+            if config["force_nnunet_model"]
+            else config["resource_urls"]["nnunet_model"][config["modality"]]
+            + "unpacked"
+        ),
     params:
         cmd_copy_inputs=get_cmd_copy_inputs,
         temp_lbl="templbl/temp.nii.gz",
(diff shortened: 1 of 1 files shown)


akhanf:
We've been using group: 'subj' on all rules so that with cluster execution, jobs are grouped so every rule from a subject is in the same job. This PR would break this, as it would require some global rule to run before a subject can run.. I've tried to avoid this so the dags for each subject aren't intertwined (ie this is why each subject has copies of template files), as it makes running a hippunfold wrapper difficult otherwise..

Wondering if the slowness you're having is just related to poor network file system performance? Is there more performant local scratch you can use to set as your output folder then copy after? Last I checked into untarring, it took 2-3 seconds when using local disk..

jordandekraker:
this should have the same grouping and be called as often as `download_nnunet_model` just before it. 

I think you're right that untarring is fast, perhaps as fast as cp, but this should still reduce my network file system load. Perhaps making a local scratch output dir would help me, but I often add more subjects as they are collected so its a bit of a pain to sync a copy. Its a good suggestion though, but this PR should still help general runtimes a hair.

akhanf:
Its the unpack_nnunet_model rule that doesn't/can't have the group: subj job grouping, and thus makes it a global rule that has to run once before any subjects

jordandekraker:
right, just like I described and again same as `download_nnunet_model`

akhanf:
Ultimately it would be good to have native windows support even if we aren't there yet.

In any case I'm still not really supportive of this. It would mean two copies exist of each model in hippunfold cache. Also the model tars are not light on number of files (that's the reason they are tarred in the first place), impinging on file number quotas that can be very restrictive on clusters. Could also be sacrificing performance when we are using  fast/local storage for the hippunfold output folder, since then every inference would need to read model data from network drives. 

If you're looking to squeeze performance likely using a local scratch (your compute nodes should have some local disk??) will make everything run alot faster. It's not a difficult workflow adjustment but overall will give you better performance especially since hippunfold has a fair bit of file i/o with small files.

Otherwise I think having this behaviour optional would be better, enabled via a flag.

jordandekraker:
its 7 files * 5 folds, i don't think its going to break the bank. Same size tarred as untarred, and it needs to travel over a network drive either way (whether to scratch first or directly to memory)

for most people this should improve performance without adding any friction. its a change so minor its not worth complicating things with an optional argument. i wasn't really expecting an argument on this but if you really hate it then i'll just forget it

jordandekraker:
also for people who use network drives like we do, it goes from read-write-read to just read. if this is one of the significant bottlenecks then we're cutting 2/3 of the traffic (assuming cpu cost to untar is negligible)

akhanf:
Ah it's only 7*2 files (!)  thats not many at all -- I had in my head there were lots of files -- then I agree and say we go all the way and remove storage of the tars altogether, and just keep the untarred folders in hippunfold_cache. 

At my computer now, so I'll take a quick look at it now to see if a simple and backwards compatible solution without symlinking is possible, otherwise can just set the tar output to temp() or something like that..

akhanf:
Symlinking the input files wasn't working (maybe related to shadow directive), but just copying the input image is pretty minimal.. If this works in your wet run tests could merge it in, should be backwards compatible (will end up replacing the tar files with directories).. 

Btw I encountered a funny snakemake issue just now where the run_inference rule was being run for both hemisphere instances, each with 12 threads, when my machine only has 12 cores.. Not sure why that was happening, maybe something broken in a recent snakemake release? Anyways, something to keep an eye on, the result in this case was one was getting killed with out of memory..

akhanf:
Also fyi untarring on my network drive took 10 seconds- is that close to what you've experienced too?

jordandekraker:
its not the untarring per se that's the issue, its copying to the output directory, then reading from the output directory to memory, then writing to the output directory again (i think). I noticed that sometimes run_inference takes a few minutes while sometimes it takes hours, even with the same cpu & RAM allocation. This sin't a problem if i run only on my local machine & local drives. I suspect its the network traffic in these read/writes, this is quite variable depending on time of day and others' usage. But I haven't broken the rule apart to test it.

akhanf:
have you confirmed that this new version indeed improves runtime? I'm wondering if this is a separate issue than just where the model sits. 

e.g. it could be related to the number of threads used for the nnunet steps prior to inference -- if too large a number, could cause lock-ups.. Right now the threads seems to be hardcoded to the number of threads in the rule, which might be too large.. I'm going to investigate this a bit more..

akhanf:
Also if it is taking hours, wondering if the snakemake scheduler is running multiple runs at once (ie problem I noticed above) causing memory issues and everything grinding to a halt..

jordandekraker:
I was hoping to confirm it in a singularity container so that i don't need to worry about my env in these nodes. 

It is very odd that it tried to use more threads than its allocation for you, but I don't think that's what is causing my issue. I have not proved this but it doesn't smell like it to me

BTW I also switched to `rsync`ing to local scratch and the run_inference and overall hippunfold runtime is way faster, but the rsync to- and from- does take quite a while, especially with all the `.snakemake/metadata`. I suspect i could delete the metadata and rely on update times instead to generate the DAG, but I'm scared to do it

akhanf:
Looking at a run with just profiling by "eye" (e.g. running htop and looking at the process while it runs), memory spikes can be pretty high, e.g. spikes of using up to half my system memory (I have 46GB total memory), and CPU usage is a bit low hovering between 6-8 cores (I have total 12 cores)..  So likely adjusting the threads parameter might speed things up a bit, but my guess is you are running into memory issues causing major slowdowns..

As a side note, one way we could have improved things is make the patch size fully encompass the image-- right now we have:

data shape: (1, 128, 256, 128)
patch size: [112 224  96]

so it's almost the same size but not quite, which means it ends up running 8 inference steps (per fold) instead of just 1, each one just shifted a little bit, so alot of redundancy. Including the 5 folds, this is 336 inferences, versus just 5. So we could have a speed up of 67X if we had done this.. 

To test this out, I just trimmed the input image to size, and the run_inference step goes from taking ~9 minutes on my 12 core machine to just over 1 minute.

If you want to try it out, e.g. can run this before inference:
```
c3d sub-1425/anat/sub-1425_hemi-L_space-corobl_desc-preproc_T1w.nii.gz -trim-to-size 96x224x112vox -o sub-1425/anat/sub-1425_hemi-L_space-corobl_desc-preproc_T1w.nii.gz
```
(note the shape is transposed)

Here is what a 1 minute inference looks like:

image

and here it is with the original size T1w overlaid to see how much is cropped..
image


It is cutting it close, but wondering it if makes sense to do this, then see if the resulting segmentation is too close to the edge, and re-run with original cropping if so..

akhanf:
our patching approach (running 8 runs with almost entirely overlapping patches) is effectively like a kind of test-time augmentation, but not a very effective one I think, so I don't think there would be much downside actually trimming the image down before inference, at least worth a shot..

akhanf:
> I was hoping to confirm it in a singularity container so that i don't need to worry about my env in these nodes.
> 
> It is very odd that it tried to use more threads than its allocation for you, but I don't think that's what is causing my issue. I have not proved this but it doesn't smell like it to me

see my below comment about memory usage, very likely the culprit especially if you're on a shared machine where available memory might go up and down depending on others.. 

> 
> BTW I also switched to `rsync`ing to local scratch and the run_inference and overall hippunfold runtime is way faster, but the rsync to- and from- does take quite a while, especially with all the `.snakemake/metadata`. I suspect i could delete the metadata and rely on update times instead to generate the DAG, but I'm scared to do it

I wouldn't rsync everything there and back each time, rather have a wrapper that just runs a subject, setting output folder to local scratch, then after the run, copy only that **subject**folder back (not the .snakemake folder). Ya you won't be able to resume runs later but overall will be alot faster..

jordandekraker:
hmm as i recall nnunet set the patch size automatically during prepare_and_preprocess, and i was scared to change it for fear of exceeding gpu memory during training (would have been in like 2019, so i'm a bit foggy on this). We can't change the patch size during inference since they need to fit into the network properly. Trimming the image to the patch size is interesting, but runs the risk of cutting off tissue, especially if `--rigid-template-reg` is turned on. We could possibly avoid this by using the "rough" segmentation we get for Dice scores to trim (in the past we did this to speed up the volumetric Laplace solver).

My run_inference times using local scratch have gone down to 2-6 mins, so i'm quite confident that network traffic was the culprit. Other rules are also exponentially faster. I just singled out run_inference to improve in this PR because its network traffic was especially high (models are 2+GB and there were 2 extra read/writes). Now that I'm down to a couple minutes of runtime using localscratch, I'm not too worried about further optimization. 

I wonder if there's a way we can make this easier for users. We can recommend in the Docs for users select a local directory as their output, and mv afterwards if needed. But I wonder if there's a snake-ier way to do this..

akhanf:
Ya patch size is set automatically based on the GPU memory size hardcoded in nnunet, so something to consider for new models -- I agree the trimmed approach would take some effort and not a huge gain since inference isn't taking that long in the first place..

For making it easy for the user to run on local disk, I would recommend building that into the run_quick script -- it already uses a tempdir customisable for the user for the input bids dataset, so wouldn't be much work to have it use for the output as well copying afterwards. I'll ask copilot to do it lol..

PR #521, opened 2025-10-01
Cleaner wet-run testing changes (move from setup-wetrun-testing)

Dhananjhay:
Same as PR #496 but cleaner


jordandekraker:
Hi DJay,

This looks great and I'd be happy to go ahead and merge, but before I do:
Is there any way we could view the png in browser instead of having to download it?

Dhananjhay:
Hey @jordandekraker,

Happy to merge this PR as well but I remember discussing with @akhanf that we might want more qc tabular results... Can always open a new PR for that. Thoughts? 

Also, seems like github has an open issue about displaying artificats in a browser: https://github.com/actions/upload-artifact/issues/14.

PR #533, opened 2026-01-28
WIP - Patch atlas syncing logic

Dhananjhay:
Swap git atlas syncing with downloading and extracting atlases from zenodo instead


Dhananjhay:
Patched in PR (#531), closing...

PR #534, opened 2026-01-31
change unfold space during construction

akhanf:
- sets the (unfoldrot) space by construction instead of at the end
- required some fixes since the unfold vol/surf spaces used in the atlas_gen rules differed from the participant-level analyses


akhanf:
~TODO: check if flip z is actually needed in the config~

Issue #548, opened 2026-05-20
v2.0.0: docker build fails with KeyError 'dseg_dentate' in import_template_dseg_dentate (--use-template-seg)


ajschadler12:
Building v2.0.0 (ffb1e85) via podman on macOS (apple silicon; using the `podman-rootful` vm template from lima).

Build fails at the 9th `/app/entrypoint.sh` call in the second `RUN` block. The prior 8 calls ran successfully.

Dockerfile line:

```shell
/app/entrypoint.sh hippunfold test_data/bids_T1w test_out participant --modality T1w --use-template-seg --use-conda --conda-create-envs-only --cores all --conda-prefix /src/conda-envs --conda-frontend mamba
```

Traceback:

```
Building DAG of jobs...
InputFunctionException in rule import_template_dseg_dentate in file "/src/hippunfold/workflow/rules/download.smk", line 142:
Error:
  KeyError: 'dseg_dentate'
Wildcards:
  subject=001
  hemi=L
Traceback:
  File "/src/hippunfold/workflow/rules/download.smk", line 149, in <lambda> (rule import_template_dseg_dentate, line 232, /src/hippunfold/workflow/rules/download.smk)
[...]

PR #550, opened 2026-05-26
Fix docker push CI

Dhananjhay:
Relevant issue #548

### DIFF
--- a/.github/workflows/push_container.yml
+++ b/.github/workflows/push_container.yml
@@ -9,6 +9,7 @@ on:
   push:
     branches:
     - dev-v2.0.0
+    - djay/fix-docker-ci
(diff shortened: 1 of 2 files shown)


akhanf:
I'm having a close look at this now -- looks like I had modified the wrong line earlier, but I think the underlying problem is that the CITI168 template doesn't have a dentate dseg defined. Other templates do though (e.g. dHCP, and MBMv3), so we can make use of the --template option to change the template here, instead of removing that line entirely (in case it misses some required conda envs when removed)..
Committing something shortly, just doing a local test first..

Dhananjhay:
Awesome! I found the same issue in issue #548.

PR #557, opened 2026-06-17
Fix push docker CI

Dhananjhay:
This PR is in regards to issue #548. Currently, the docker build is failing because gh servers are running out of space during the process.


akhanf (approved):
Thanks for sorting this out!

Issue #564, opened 2026-08-23
hippunfold command missing from container in dev-v2.0.0 (only hippunfold-quick is exposed)


RicardoRios46:
The hippunfold main CLI command is not available inside the khanlab/hippunfold:dev-v2.0.0 container, while hippunfold-quick is present. This cause me some issues while trying to run the hippunfold container as a containered rules inside my snakemake workflow.

Steps to reproduce
```bash
apptainer pull docker://khanlab/hippunfold:dev-v2.0.0
apptainer shell hippunfold_dev-v2.0.0.sif
Apptainer> which hippunfold
Apptainer> which hippunfold-quick
/usr/local/bin/hippunfold-quick
```

No output for hippunfold (command not found).

PR #565, opened 2026-08-23
Add docker wrapper for hippunfold command to be exposed to apptainer

RicardoRios46:
suggested  fix  for #564 

Claude helped me with this one. Apparently "apptainer exec" bypass the image entrypoints created by pixi entirely. Copying the wrapper for hippunfold-quick (which is working fine) should expose hippunfold to the apptainer exec too.

claude explanation in case it helps:
"Because ENTRYPOINT ["/app/entrypoint.sh"] is what activates the pixi environment (via pixi shell-hook), hippunfold only resolves on $PATH when the container is run through its entrypoint (docker run/apptainer run). Any invocation that bypasses the entrypoint — notably apptainer shell / apptainer exec, common for interactive debugging — leaves the pixi env unactivated, so hippunfold, python3, and pip are all missing from $PATH, while hippunfold-quick still works because it has its own static wrapper."


RicardoRios46:
The following command correclty outputs the help

```
apptainer exec hippunfold_dev-v2.0.0.sif /app/entrypoint.sh /src/hippunfold/run.py --help
```

Note: the first commit was using "hippunfuld" instead of poiting to "/src/hippunfold/run.py" which also seems to work. I am using run.py to match the quick wrapper.

PR #567, opened 2026-09-11
Fix checkpoint and trainer values for ADNI_T1w_v1

akhanf:
This pull request updates the configuration for the `nnunet_models` section in `hippunfold/config/snakebids.yml` to correct the values for the `checkpoint` and `trainer` fields for the `adniT1w_v1` model.

Model configuration corrections:

* Changed the `checkpoint` value from `model_best` to `nnUNetTrainer` for the `adniT1w_v1` model.
* Changed the `trainer` value from `nnUNetTrainer` to `Task301_ADNI_T1w_successful` for the `adniT1w_v1` model.

Note: these seem to have been typos when originally training the model @Bradley-Karat, or maybe some files were manually renamed after training? 

This became a bug when we moved to explicit declaration of these values (which was done to support both v1 and v2 nnunet), rather than parsing them from the filename parts.


Bradley-Karat:
I can't remember specifically, but likely that I manually renamed some files after training

